Issue #302, opened 2024-10-13
Store snapshot debt as a 18 decimal value


smol-ninja:
As per [cantina finding 11](https://cantina.xyz/code/99ae802b-f05c-4e36-a1d1-240d5146649c/findings/11), storing snapshot debt as a 18 decimal value can increase precision.

Note: use `uint256` for `snapshotDebt` as per https://github.com/sablier-labs/flow/discussions/297#discussioncomment-10935202.

Issue #307, opened 2024-10-14
replace the delay logic from invariant with an inequality


smol-ninja:
As discussed in https://github.com/sablier-labs/flow/issues/289, remove the delay logic from `invariant_TotalStreamedWithDelayEqTotalDebtPlusWithdrawn` and use inequality as follows:

- approximately assert that "totalStreamedAmount = flow.totalDebtOf(streamId) + flowStore.withdrawnAmounts(streamId)" with a delta of 10 wei.

PR #312, opened 2024-10-15
refactor: sd as 18 decimals

andreivladbrg:
Closes https://github.com/sablier-labs/flow/issues/302

#### Changes

- store the snapshot debt as 18 decimals 
- rename `ongoingDebt` to `ongoingDebtScaled`
    - use only `uint256` type in this function
- rename `snapshotDebt` to `snapshotDebtScaled`
- update `withdraw` function accordingly
- reintroduce the scale/descale helpers functions
- update `depletionTime` function
    - the biggest change here is that we need to add to the stream balance with scaled `1` value (i.e. for USDC it would be`1e12`)
- use `uint256` in scale/descale function from tests
    - the reason for this is that it safer to up cast from `uint128` to `uint256` compared to `uint256` to `uint128`
- add `ONE_MONTH_DEBT_18D` constant for `snapshotDebt` tests


andreivladbrg:
@smol-ninja nevermind for now the fail inv test, we will change the `eq` to `ineq` as we decided in https://github.com/sablier-labs/flow/issues/307

src/libraries/Helpers.sol
```diff
@@ -49,4 +49,34 @@ library Helpers {
         // Calculate the broker fee amount that is going to be transferred to the `broker.account`.
         (brokerFeeAmount, depositAmount) = calculateAmountsFromFee(totalAmount, broker.fee);
     }
+
+    /// @notice Descales the provided `amount` to be denoted in the token's decimals.
```
smol-ninja, line 53:
```suggestion
    /// @notice Descales the provided `amount` from 18 decimals fixed-point number to token's decimals fixed-point number.
```

src/libraries/Helpers.sol
```diff
@@ -49,4 +49,34 @@ library Helpers {
...
+    /// @dev The following logic is used to denormalize the amount:
+    /// - If the token has exactly 18 decimals, the amount is returned as is.
+    /// - if the token has fewer than 18 decimals, the amount is divided by $10^(18 - tokenDecimals)$.
+    function descaleAmount(uint256 amount, uint8 decimals) internal pure returns (uint256) {
+        if (decimals > 18) {
```
smol-ninja, line 58: No need to have `>` in this. Otherwise it goes against the natspecs and we would have to descale amount to 18 decimals if `decimals > 18`

```suggestion
        if (decimals == 18) {
```

src/libraries/Helpers.sol
```diff
@@ -49,4 +49,34 @@ library Helpers {
         // Calculate the broker fee amount that is going to be transferred to the `broker.account`.
         (brokerFeeAmount, depositAmount) = calculateAmountsFromFee(totalAmount, broker.fee);
     }
+
+    /// @notice Descales the provided `amount` to be denoted in the token's decimals.
+    /// @dev The following logic is used to denormalize the amount:
```
smol-ninja, line 54: We can remove `dev` comment. Instead of that how about adding comments inside the logic i.e. over `if` and over `unchecked`.

src/libraries/Helpers.sol
```diff
@@ -49,4 +49,34 @@ library Helpers {
...
+            uint256 scaleFactor = 10 ** (18 - decimals);
+            return amount / scaleFactor;
+        }
+    }
+
+    /// @notice Scales the provided `amount` to be denoted in 18 decimals.
+    /// @dev The following logic is used to normalize the amount:
+    /// - If the token has exactly 18 decimals, the amount is returned as is.
+    /// - If the token has fewer than 18 decimals, the amount is multiplied by $10^(18 - tokenDecimals)$.
+    function scaleAmount(uint256 amount, uint8 decimals) internal pure returns (uint256) {
```
smol-ninja, line 72: Same suggestions as above.

src/SablierFlow.sol
```diff
@@ -69,35 +69,35 @@ contract SablierFlow is
...
-            return _streams[streamId].snapshotTime + solvencyPeriod;
+
+            depletionTime = _streams[streamId].snapshotTime + solvencyPeriod;
         }
     }
 
     /// @inheritdoc ISablierFlow
     function ongoingDebtOf(uint256 streamId) external view override notNull(streamId) returns (uint256 ongoingDebt) {
-        ongoingDebt = _ongoingDebtOf(streamId);
+        ongoingDebt = Helpers.descaleAmount({
```
smol-ninja, line 97: Whats the point of returning `ongoingDebt` as descaled but `snapshotDebt` as scaled? We should return `ongoingDebt` as scaled version so that the following holds true.

`Helpers.descale(ongoingDebt + snapshotDebt) == totalDebt`

Otherwise, it can return false. Because if  `ongoingDebt` is descaled, it would introduce precision error to above calculations.

And for this to work, no need to change `ongoingDebt` to `scaledOngoingDebt` for public function. Similar to how we mention that `getSnapshotDebt` returns value in 18-decimal format, we can do the same for `ongoingDebt` function.

src/SablierFlow.sol
```diff
@@ -449,11 +449,11 @@ contract SablierFlow is
...
     }
 
-    /// @dev Calculates the ongoing debt accrued since last snapshot. Return 0 if the stream is paused or
-    /// `block.timestamp` is less than or equal to snapshot time.
-    function _ongoingDebtOf(uint256 streamId) internal view returns (uint256 ongoingDebt) {
-        uint40 blockTimestamp = uint40(block.timestamp);
-        uint40 snapshotTime = _streams[streamId].snapshotTime;
+    /// @dev Calculates the ongoing debt, as a 18-decimals fixed point number, accrued since last snapshot. Return 0 if
+    /// the stream is paused or `block.timestamp` is less than or equal to snapshot time.
+    function _scaledOngoingDebtOf(uint256 streamId) internal view returns (uint256) {
```
smol-ninja, line 454: If we agree to returning 18 decimal value in `ongoingDebt`, we can keep it as `_scaledOngoingDebtOf`

smol-ninja (changes requested):

src/SablierFlow.sol
andreivladbrg: > Whats the point of returning ongoingDebt as descaled but snapshotDebt as scaled?

i had these three reasons in mind: 
- consistency over "debt" calculation functions
- no point to return an amount of _calculated_ "debt" _externally_ that cannot transferred
- Razvan made a point to have them return only in token's decimals IIRC (will ask again)

but this makes sense and i tend to agree:
> as scaled version so that the following sanity check holds true.
> Helpers.descale(ongoingDebt + snapshotDebt) == totalDebt

smol-ninja: Thanks for sharing the reasons. Just a comment below.

> no point to return an amount of calculated "debt" externally that cannot transferred

Since we are returning `snapshotDebt` in 18 decimal already, its not true anymore. An external actor may not have this information that `ongoingDebt` is being calculated during runtime whereas `snapshotDebt` is a stored in slots. So imo the consistency is when we return all debts in the same decimal format including totalDebt. But then it creates a poor DX because `totalDebt` and `balance` are required to be compared to understand the stream's solvency.

Now, since `snapshotDebt` and `ongoingDebt` are like two saiyans and both of them are required to form a super saiyan called `totalDebt`, it would be a good idea to return `snapshotDebt` and `ongoingDebt` in 18 decimal atleast so people know these two are linked to each other.

andreivladbrg: as discussed on slack, we will rename the function to `ongoingDebtScaled`, and the `snapshotDebt` to `snapshotDebtScaled`

smol-ninja: and variables as well.

> What if we rename `snapshotDebt` to `snapshotDebtScaled` and `ongoingDebt` to `ongoingDebtScaled` everywhere and provide no function that return descaled versions of these two, because if we do, then it again bring in the same confusion that equality wont hold true between these two and `totalDebt`.

- Always use `ongoingDebtScaled` and `snapshotDebtScaled` names when referring to ongoing debt and snapshot debt related functions and variables to represent them in 18-decimals fixed-point numbers
- Don't offer descaled getters for them to avoid any confusion to users.

cc @razgraf

smol-ninja (changes requested):
Thank you for making all these changes. Clearly this is a big PR and has more impacts than I imagined. A few comments:

- I can see two variations of `scaled` naming: `scaledABC` and `abcScaled`. I suppose you are using `abcScaled` for function names (`ongoingDebtScaledOf` and `getSnapshotDebtScaled`) whereas `scaledABC` for variables names. If so, there are inconsistencies at a few places such as use of `snapshotDebtScaled` variable in `getSnapshotDebtScaled` function. Lets stick to one convention so that there is no confusion when deciding for a variable names that requires 18 decimal format. 

My suggestion would be to either use `scaledABC` format or `ABCScaled` format for all variables and function naming and not both. Whats your preference?

- Would it be more consistent if we use the same naming convention for snapshot debt getter as well i.e. `snapshotDebtOf` similar to `ongoingDebtOf`? Its the only function that looks out of touch since it leans more towards getters defined in `ISablierFlow` than `ISablierFlowBase`. If you are aligned, we can consider refactoring it in the next release of Flow.

smol-ninja:
One more comment: Technical doc is now outdated given that `sd` has been refactored to follow 18 decimal format. What do we do about it now? Here are some of the ideas:

- Lets remove numbers from it as it would require to be re-calculated.
- Only adds theory and formula as its easy to do.

Wdyt? if you think you want it to have numbers as well then we will have to update them as well as charts (I suppose charts are outdated too right?)

andreivladbrg:
> * can see two variations of `scaled` naming: `scaledABC` and `abcScaled`

actually, my intention was to only have one, and that is `abcScaled`. the only exception, as you mentioned, is `ongoingDebtOf`, which would indeed end up in a situation where the function would have a sounded weird, "ongoingDebtOfScaled".

> My suggestion would be to either use scaledABC format or ABCScaled format for all variables and function naming and not both. Whats your preference?

ofc, we should have just one, as said above. how would you keep consistency and still have a good function name? my preference is `abcScaled`

> snapshotDebtOf similar to ongoingDebtOf

i think that would be misleading. snapshot debt is a state variable, while the other is a function that calculates an amount in real time. we should definitely keep the "get" keyword for all getters, as it only reads a state, and having "Of" points out that it’s different from the classic getters (dynamic function)

smol-ninja:
> actually, my intention was to only have one, and that is abcScaled. 

Got it. Then lets use `abcScaled` everywhere, and for ongoing debt, we can use `ongoingDebtScaledOf`.

tests/integration/concrete/pause/pause.t.sol
```diff
@@ -62,19 +62,19 @@ contract Pause_Integration_Concrete_Test is Integration_Test {
...
 
         // It should pause the stream.
-        test_Pause();
+        _test_Pause();
     }
 
-    function test_Pause() internal {
-        uint256 initialTotalDebt = flow.totalDebtOf(defaultStreamId);
+    function _test_Pause() private {
+        uint256 expectedSnapshotDebt = flow.ongoingDebtScaledOf(defaultStreamId);
```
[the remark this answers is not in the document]
andreivladbrg: > Wouldn't the correct approach here be to assign expectedSnapshotDebt to flow.getSnapshotDebtScaled(streamID) + flow.ongoingDebtScaledOf(streamId) since after pause, sd would get incremented by od?

techincally yes, but this a _concrete_ test, we should have actually only "contrete" values to test them, since the `sd` is now normalized we can use `ONE_MONTH_DEBT_18D`

your approach is ideal in an inv/fuzz environment IMO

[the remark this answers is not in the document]
smol-ninja: But you did not use `ONE_MONTH_DEBT_18D` instead `flow.ongoingDebtScaledOf`.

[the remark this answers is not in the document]
andreivladbrg: Yes, mb, I should change this, will push

tests/integration/concrete/total-debt-of/totalDebtOf.t.sol
```diff
@@ -14,27 +14,24 @@ contract TotalDebtOf_Integration_Concrete_Test is Integration_Test {
     function test_GivenPaused() external givenNotNull {
         flow.pause(defaultStreamId);
 
-        uint256 snapshotDebt = flow.getSnapshotDebt(defaultStreamId);
-        uint256 totalDebt = flow.totalDebtOf(defaultStreamId);
+        uint256 snapshotDebtScaled = flow.getSnapshotDebtScaled(defaultStreamId);
 
-        assertEq(totalDebt, snapshotDebt, "total debt");
+        assertEq(ONE_MONTH_DEBT_18D, snapshotDebtScaled, "total debt");
```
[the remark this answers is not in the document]
andreivladbrg: correct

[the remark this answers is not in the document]
smol-ninja: "it" implies the function that the test is written for. In this case, its `totalDebtOf`

> it should return snapshot debt

This means that `totalDebtOf` should return snapshot debt.

> it should return the sum of snapshot debt and ongoing debt

This means that `totalDebtOf` should return snapshot debt + ongoing debt.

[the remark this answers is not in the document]
andreivladbrg: updated in 663e9e83b41e3d206a64cb5e1d40ce37adc8cb65

tests/integration/concrete/pause/pause.t.sol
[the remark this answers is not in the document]
andreivladbrg: 663e9e83b41e3d206a64cb5e1d40ce37adc8cb65

src/SablierFlow.sol
```diff
@@ -69,35 +69,38 @@ contract SablierFlow is
             return 0;
         }
 
-        uint256 snapshotDebt = _streams[streamId].snapshotDebt;
+        uint8 tokenDecimals = _streams[streamId].tokenDecimals;
+        uint256 scaledBalance = Helpers.scaleAmount({ amount: balance, decimals: tokenDecimals });
```
smol-ninja, line 73: `balanceScaled`

src/SablierFlow.sol
```diff
@@ -470,35 +473,20 @@ contract SablierFlow is
...
 
-    /// @notice Calculates the total debt.
-    /// @dev The total debt is the sum of the snapshot debt and the ongoing debt. This value is independent of the
-    /// stream's balance.
+    /// @dev The total debt is the sum of the snapshot debt and the ongoing debt descaled to token's decimal. This
+    /// value is independent of the stream's balance.
     function _totalDebtOf(uint256 streamId) internal view returns (uint256) {
-        // Calculate the total debt.
-        return _streams[streamId].snapshotDebt + _ongoingDebtOf(streamId);
+        uint256 scaledTotalDebt = _ongoingDebtScaledOf(streamId) + _streams[streamId].snapshotDebtScaled;
```
smol-ninja, line 488: `totalDebtScaled`

src/SablierFlow.sol
```diff
@@ -525,12 +513,12 @@ contract SablierFlow is
             revert Errors.SablierFlow_RatePerSecondNotDifferent(streamId, newRatePerSecond);
         }
 
-        uint256 ongoingDebt = _ongoingDebtOf(streamId);
+        uint256 scaledOngoingDebt = _ongoingDebtScaledOf(streamId);
```
smol-ninja, line 516: `ongoingDebtScaled`

src/SablierFlow.sol
```diff
@@ -792,17 +784,20 @@ contract SablierFlow is
             revert Errors.SablierFlow_Overdraw(streamId, amount, withdrawableAmount);
         }
 
+        // Calculate the amount scaled.
+        uint256 scaledAmount = Helpers.scaleAmount(amount, tokenDecimals);
```
smol-ninja, line 788: `amountScaled`

tests/integration/concrete/ongoing-debt-of/ongoingDebtOf.t.sol
smol-ninja, line 1: I renamed it to `ongoingDebtScaledOf.t.sol` in my commit which has now been overridden so it has been reverted back. It should be `ongoingDebtScaledOf.t.sol`

tests/integration/concrete/ongoing-debt-of/ongoingDebtOf.tree
smol-ninja, line 1: same renaming of file here.

tests/integration/concrete/total-debt-of/totalDebtOf.tree
```diff
@@ -3,7 +3,7 @@ TotalDebtOf_Integration_Concrete_Test
 │  └── it should revert
 └── given not null
    ├── given paused
-   │  └── it should return snapshot debt
+   │  └── it should return the
```
smol-ninja, line 6: error here.

tests/integration/fuzz/ongoingDebtOf.t.sol
smol-ninja, line 1: file name here.

smol-ninja (changes requested):
Last few suggestions then looks good here.

smol-ninja (approved):
Fanstatic. Less goooo 🚀

Issue #300, opened 2024-10-13
include `sender` and `recipient` in deposit


smol-ninja:
As per [Cantina finding 7](https://cantina.xyz/code/99ae802b-f05c-4e36-a1d1-240d5146649c/findings/7), deposits are susceptible of reorg attacks. Thus, include `sender` and `recipient` in `deposit`'s input parameters. This would avoid stealing funds from depositors during chain re-orgs.

PR #313, opened 2024-10-15
refactor: deposit input parameters

andreivladbrg:
Depends on https://github.com/sablier-labs/flow/pull/312 

Closes https://github.com/sablier-labs/flow/issues/300

**Note:** The reason for not including `sender` and `recipient` parameters in `depositAndPause` and `restartAndDeposit` is that only the stream's sender can call these functions. Therefore, it is implicitly protected from reorg attacks. If we add any check, it would only be for the recipient. Let me know if you find that useful.

### DIFF
--- a/src/SablierFlow.sol
+++ b/src/SablierFlow.sol
@@ -242,13 +242,16 @@ contract SablierFlow is
     /// @inheritdoc ISablierFlow
     function deposit(
         uint256 streamId,
-        uint128 amount
+        uint128 amount,
+        address sender,
+        address recipient
     )
         external
         override
         noDelegateCall
         notNull(streamId)
         notVoided(streamId)
+        notDifferentActors(streamId, sender, recipient)
         updateMetadata(streamId)
     {
         // Checks, Effects, and Interactions: deposit on stream.
@@ -279,6 +282,8 @@ contract SablierFlow is
     function depositViaBroker(
         uint256 streamId,
         uint128 totalAmount,
+        address sender,
+        address recipient,
         Broker calldata broker
     )
         external
@@ -286,6 +291,7 @@ contract SablierFlow is
         noDelegateCall
         notNull(streamId)
         notVoided(streamId)
+        notDifferentActors(streamId, sender, recipient)
         updateMetadata(streamId)
     {
         // Checks, Effects, and Interactions: deposit on stream through broker.
--- a/src/SablierFlow.sol
+++ b/src/SablierFlow.sol
@@ -251,9 +251,11 @@ contract SablierFlow is
         noDelegateCall
         notNull(streamId)
         notVoided(streamId)
-        notDifferentActors(streamId, sender, recipient)
         updateMetadata(streamId)
     {
+        // Check: the provided sender and recipient match the stream's sender and recipient.
+        _verifyStreamSenderRecipient(streamId, sender, recipient);
+
         // Checks, Effects, and Interactions: deposit on stream.
         _deposit(streamId, amount);
     }
@@ -291,9 +293,11 @@ contract SablierFlow is
         noDelegateCall
         notNull(streamId)
         notVoided(streamId)
-        notDifferentActors(streamId, sender, recipient)
         updateMetadata(streamId)
     {
+        // Check: the provided sender and recipient match the stream's sender and recipient.
+        _verifyStreamSenderRecipient(streamId, sender, recipient);
+
         // Checks, Effects, and Interactions: deposit on stream through broker.
         _depositViaBroker(streamId, totalAmount, broker);
     }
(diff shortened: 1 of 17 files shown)


andreivladbrg:
@smol-ninja updated the branch with sd to 18 change

src/abstracts/SablierFlowBase.sol
```diff
@@ -66,6 +66,21 @@ abstract contract SablierFlowBase is
                                       MODIFIERS
     //////////////////////////////////////////////////////////////////////////*/
 
+    /// @dev Checks whether the provided addresses matches stream's sender and recipient.
+    modifier notDifferentActors(uint256 streamId, address sender, address recipient) {
```
smol-ninja, line 70: How about renaming it to `checkParams`?

and revert with a single error if any of the checks fails: `revert Errors.InvalidDepositParams(sender, recipient)`

src/interfaces/ISablierFlow.sol
```diff
@@ -245,10 +245,14 @@ interface ISablierFlow is
     /// - Must not be delegate called.
     /// - `streamId` must not reference a null or a voided stream.
     /// - `amount` must be greater than zero.
+    /// - `sender` and `recipient` must match the stream's sender and recipient addresses.
     ///
     /// @param streamId The ID of the stream to deposit to.
     /// @param amount The deposit amount, denoted in token's decimals.
-    function deposit(uint256 streamId, uint128 amount) external;
+    /// @param amount The deposit amount, in token decimals.
```
smol-ninja, line 252: Duplicate.

src/SablierFlow.sol
```diff
@@ -242,13 +242,16 @@ contract SablierFlow is
     /// @inheritdoc ISablierFlow
     function deposit(
         uint256 streamId,
-        uint128 amount
+        uint128 amount,
+        address sender,
```
smol-ninja, line 246: Should we rename it to `streamSender` and `streamRecipient` to signal that these are input parameters?

andreivladbrg: > Eww. The deposit function has started to look ugly. :\

yeah… it does unfortunetely... what can we do about it?

> Should we rename it to streamSender and streamRecipient to signal that these are input parameters?

i would say no, for the same reason we don’t name the parameter `depositAmount`, but just `amount`. since we pass the `streamId`, I think it implicitly means we are talking about the stream’s sender/recipient (the natspec also says so)

src/interfaces/ISablierFlow.sol
andreivladbrg: correct

src/SablierFlow.sol
smol-ninja: > what can we do about it?

Encode it and pass it as bytes, like you previously suggested? Then rename the input parameter to `configuration`. What do you think about that?

Imagine `sender` making a deposit and confused as to why he has to supply `sender` address when he is the `msg.sender`. 

Wdyt? I am just suggesting as it look weird to me.

src/abstracts/SablierFlowBase.sol
smol-ninja: @PaulRBerg you have any suggestion here does it look good?

src/SablierFlow.sol
smol-ninja: @PaulRBerg here as well.

smol-ninja: Lets keep it this way as per the comment from Slack: https://sablier.slack.com/archives/CT3NXSNMS/p1729114008318439?thread_ts=1729113479.102569&cid=CT3NXSNMS

Variable naming looks good too.

src/abstracts/SablierFlowBase.sol
smol-ninja: How about the following name @andreivladbrg:
- `areStreamParties(sender, recipient)` similar to other naming convention we have such as `notNull`, `notPaused`, `isTransferrable` etc.

PaulRBerg: Naming is difficult!

I don't like `notDifferentActors`.

What about `onlyStreamUsers` or `onlyStreamParties`?

smol-ninja: `onlyStreamParties` does not sound clear as well. It may mean if only stream parties are allowed to call the function. What do you think about `_verifyStreamSenderRecipient`? and Instead of modifier, make it an internal function since its only used in two functions: `deposit` and `depositViaBroker`. It is not going well among other modifiers given that it has got three inputs.

PaulRBerg: Good point.

And yeah, let's go with an internal function.

It really is a small design choice in the grand scheme of things — let's get hasty.

smol-ninja (approved):

PR #314, opened 2024-10-16
docs: refactor technical doc

smol-ninja:
Refactors technical doc in line with https://github.com/sablier-labs/flow/pull/312.


TECHNICAL-DOC.md
```diff
@@ -13,11 +13,11 @@ pause it or void it at a later date.
 A stream is represented by a struct, which can be found in
 [`DataTypes.sol`](https://github.com/sablier-labs/flow/blob/ba1c9ba64907200c82ccfaeaa6ab91f6229c433d/src/types/DataTypes.sol#L41-L76).
 
-The debt is tracked using `snapshotDebtScaled` and `snapshotTime`. At snapshot, the following events are taking place:
+The debt is tracked using "snaoshot debt" and "snapshot time". At snapshot, the following events are taking place:
```
PaulRBerg, line 16: typo

andreivladbrg (changes requested):
> the unlock interval takes into account the last withdraw time ($wt$) and not snapshot time ($st$) anymore. Thats what the PR is about.

I don’t understand this very well. Can you please rephrase it? 
As far as I can see, I think you wanted to change how we refer to the moment in time `t`. Changing `st` to `wt` is not precisely correct. For example, these constant intervals are not correct, as the constant interval starts from st up to the moment the withdrawal happened:

https://github.com/sablier-labs/flow/blob/cf2a59f9015314ae2533bf5e0f16fb2ad54e9964/TECHNICAL-DOC.md?plain=1#L393-L395


> Without changing the numbers we can say that time values are considered since last withdraw time and not the last snapshot time

The number shouldn’t be changed—the delay in the current example (`rps = 0.000000_011574e18`) should remain the same.

> This will require updating the charts with the new labels

correct, will do

> This is due to the fact that with numbers and charts, the cost of maintenance of the technical doc is non-zero. The purpose of the doc is to explain delays and precision issues in calculations of withdrawn amount which imo is already well explained. 

The fact that the maintenance is non-zero is correct. I believe having concrete examples and clear graphs in our theory section will help external readers understand the delay issue more quickly and effectively. For us, it took a while to precisely understand what was actually happening, so I believe the cost is worth it to minimize future questions about the technical part.

TECHNICAL-DOC.md
```diff
@@ -448,19 +451,20 @@ be calculated as:
...
 
 We can also reverse engineer the delay from the _rescaled_ ongoing debt:
 
 ```math
 \begin{aligned}
-\text{od} &= \frac{rps \cdot (t - \text{st})}{sf} \\
-R_\text{od} &= od \cdot sf \\
-delay &= t - st - \frac{R_\text{od}}{rps} - 1 \\
+\text{td} &= od + rps \cdot (t - \text{st}) \\
+delay &= t - wt - \frac{R_\text{td}}{rps} - 1 \\
```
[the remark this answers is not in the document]
smol-ninja: Yeah agree to remove it.

TECHNICAL-DOC.md
```diff
@@ -13,11 +13,11 @@ pause it or void it at a later date.
...
 [`DataTypes.sol`](https://github.com/sablier-labs/flow/blob/ba1c9ba64907200c82ccfaeaa6ab91f6229c433d/src/types/DataTypes.sol#L41-L76).
 
-The debt is tracked using `snapshotDebtScaled` and `snapshotTime`. At snapshot, the following events are taking place:
+The debt is tracked using "snapshot debt" and "snapshot time". At snapshot, the following events are taking place:
 
-1. `snapshotDebtScaled` is incremented by `ongoingDebtScaled` where
-   `ongoingDebtScaled = rps * (block.timestamp - snapshotTime)`.
-2. `snapshotTime` is updated to `block.timestamp`.
+1. snapshot debt is incremented by ongoing debt where
+   $\text{ongoing debt} = rps \times (\text{block timestamp} - \text{snapshot time})$.
```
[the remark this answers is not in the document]
smol-ninja: sounds good.

smol-ninja:
@andreivladbrg do you remember what we discussed about it last week? I forgot to make notes and now can't recall what we decided about this doc.

smol-ninja:
@andreivladbrg I have made the suggested changes. Let me know if it looks good now.

andreivladbrg:
@smol-ninja as discussed on slack, we should remove everything from this line and below https://github.com/sablier-labs/flow/blob/4c198886f80a472aec17f3a54b55830f1a29e961/TECHNICAL-DOC.md?plain=1#L215

since there is no more delay due to sd in 18 decimals, also it can be proved by these equality checks in the PR: https://github.com/sablier-labs/flow/pull/320

smol-ninja:
Not that line. MVT is one of the two problems that led us to use 18-decimal fixed-point number for rps and sd. Everything [from here](https://github.com/sablier-labs/flow/blob/docs/technical-doc/TECHNICAL-DOC.md#delay-due-to-descaling) should be deleted.

tests/integration/concrete/withdraw-delay/withdrawDelay.t.sol
```diff
@@ -5,8 +5,7 @@ import { ud21x18 } from "@prb/math/src/UD21x18.sol";
 
 import { Integration_Test } from "../../Integration.t.sol";
 
-/// @dev A series of demonstrative tests that help to understand the precision and delay problem detailed in
-/// `precision.md` file.
+/// @dev A series of demonstrative tests that help to understand prove the absence of delay in the withdrawal process.
```
smol-ninja, line 8: If something is not present, why should we have tests to prove its absence? Also the equality in https://github.com/sablier-labs/flow/pull/320 is sufficient to prove that streaming works as expected.

TECHNICAL-DOC.md
```diff
@@ -223,248 +224,8 @@ By using an 18-decimal `rps` system, we can allow streaming of amount less than
```
smol-ninja, line 224: lets add a subheading to distinguish it from the problems section.

### Conclusion

or 

### Solution

andreivladbrg: 7a2693ea8ff8ecb3fa69fba6bb970c73d870d0d0

tests/integration/concrete/withdraw-delay/withdrawDelay.t.sol
andreivladbrg: I am thinking of a way to mention these in the docs
but not yet sure, so i would temporarily keep them

wdyt?

smol-ninja: We started with an implementation that had these problems and required us to explain it. And then we solved them. So if you think from a fresh perspective, this problem does not exist.

So for someone looking at the code with fresh mind, the explanations in technical doc and the assumption in security is sufficient. I think this contract is no longer required.

andreivladbrg: fair enough

andreivladbrg: aa97ca0b087d9f93c36f1f4f3798a31caccf0562

smol-ninja (commented):
looks good to me now.

andreivladbrg (approved):
lessss gooo, no more delay 🚀🚀

PR #315, opened 2024-10-16
test: inequality in streamed amount invariant

andreivladbrg:
Closes https://github.com/sablier-labs/flow/issues/307

### DIFF
--- a/tests/invariant/Flow.t.sol
+++ b/tests/invariant/Flow.t.sol
@@ -342,17 +342,8 @@ contract Flow_Invariant_Test is Base_Test {
 
             // Skip the voided streams.
             if (!flow.isVoided(streamId)) {
-                (uint256 totalStreamedAmount, uint256 totalStreamedAmountWithDelay) =
-                    calculateTotalStreamedAmounts(flowStore.streamIds(i), flow.getTokenDecimals(streamId));
-
-                assertGe(
-                    totalStreamedAmount,
-                    totalStreamedAmountWithDelay,
-                    "Invariant violation: total streamed amount without delay >= total streamed amount with delay"
-                );
-
-                assertEq(
-                    totalStreamedAmountWithDelay,
+                assertLe(
+                    calculateTotalStreamedAmount(flowStore.streamIds(i), flow.getTokenDecimals(streamId)),
                     flow.totalDebtOf(streamId) + flowStore.withdrawnAmounts(streamId),
                     "Invariant violation: total streamed amount with delay = total debt + withdrawn amount"
                 );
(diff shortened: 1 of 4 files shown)


tests/invariant/Flow.t.sol
```diff
@@ -351,10 +351,11 @@ contract Flow_Invariant_Test is Base_Test {
...
 
-                assertEq(
-                    totalStreamedAmountWithDelay,
-                    flow.totalDebtOf(streamId) + flowStore.withdrawnAmounts(streamId),
-                    "Invariant violation: total streamed amount with delay = total debt + withdrawn amount"
+                // solhint-disable max-line-length
+                assertLe(
+                    (flow.totalDebtOf(streamId) + flowStore.withdrawnAmounts(streamId)) - totalStreamedAmountWithDelay,
+                    3,
+                    "Invariant violation: total debt + withdrawn amount must be at most 3 units greater than total streamed amount with delay"
```
smol-ninja, line 358: Why 3? I suppose it was failing for 2? If so, lets change it to 10 because it can fail at 3 as well. 

Also, as mentioned in [issue](https://github.com/sablier-labs/flow/issues/307), why not compare `totalStreamedAmount` with `flow.totalDebtOf(streamId) + flowStore.withdrawnAmounts(streamId)` instead of `totalStreamedAmountWithDelay`?

The purpose of `totalStreamedAmountWithDelay` was to not have this inequality but now since we are moving towards inequality, it would be more accurate invariant to confirm that actual streamed is within 10 wei of total streamed as a good assumption.

Because if we compare `totalStreamedAmountWithDelay` with `totalDebtOf + withdrawnAmounts` which is actual streamed, the deviation from `totalStreamed` can be bigger than the delta of 3.

andreivladbrg: > Why 3? I suppose it was failing for 2? If so, lets change it to 10 because it can fail at 3 as well.

nope, it was not failing for 2. all the CI invariants that failed (for those I found) had only a difference of 1. it uses the amount calculated including the delay, which theoretically should be equal. the reason for the failure is that we take the delay from the $t$ moment instead of calculating it from $t_0$ to $t$, which completely shifts it to the right

tests/invariant/Flow.t.sol
```diff
@@ -342,17 +342,8 @@ contract Flow_Invariant_Test is Base_Test {
...
-
-                assertGe(
-                    totalStreamedAmount,
-                    totalStreamedAmountWithDelay,
-                    "Invariant violation: total streamed amount without delay >= total streamed amount with delay"
-                );
-
-                assertEq(
-                    totalStreamedAmountWithDelay,
+                assertLe(
```
smol-ninja, line 354: I wonder why is there an "`Le`" here? Previously, we had the following relations:
1. $totalStreamedAmount \ge totalStreamedAmountWithDelay$
2. $totalStreamedAmountWithDelay = totalDebt + totalWithdrawn$

Therefore from the above two, the new relation should be
- $totalStreamedAmount \ge totalDebt + totalWithdrawn$

and not the opposite.

smol-ninja (changes requested):

smol-ninja (approved):

tests/invariant/Flow.t.sol
smol-ninja: https://github.com/sablier-labs/flow/pull/315/commits/45436e9cbf3ff149f5f3f8f7bc059b91ffdaaa94

PR #317, opened 2024-10-17
fix: depletionTimeOf calculation

smol-ninja:
There was a small discrepancy between the definition of "depletion timestamp" and its implementation.

Depletion time is defined as the UNIX timestamp at which the total debt exceeds stream balance by 1 wei. That is, at depletion time, $\text{total debt} = \text{stream balance} + 1$. This also means that at depletion time, $\text{uncovered debt} = 1$. At depletion time, the streamed accrues 1 wei of uncovered debt. Therefore, this PR closes the gap between the definition and the implementation.

Therefore, consider the following cases and the expected results:

### Changelog
- Adds clarity to the definition of `depletionTimeOf` function
- Adds `+ 1` to the returned value of `depletionTimeOf` if division is not exact
- Compare values 1 second before, at, and 1 second after depletion timestamp in fuzz tests

### DIFF
--- a/src/SablierFlow.sol
+++ b/src/SablierFlow.sol
@@ -74,21 +74,30 @@ contract SablierFlow is
 
         uint256 snapshotDebtScaled = _streams[streamId].snapshotDebtScaled;
 
-        // If the stream has uncovered debt, return zero.
-        if (snapshotDebtScaled + _ongoingDebtScaledOf(streamId) > balanceScaled) {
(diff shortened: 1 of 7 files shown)


smol-ninja:
RustyRabbit has given a green signal. @PaulRBerg if you can have a quick look and approve, I can merge and prepare it for Codehawk. Thank you.

PaulRBerg (approved):
LGTM, let's just add an explanatory comment in which we define MVT

PR #320, opened 2024-10-24
test: use equality in the withdraw multiple test and in invariant

andreivladbrg:
As we changed the `sd` to 18 decimals, I discovered that the delay explained in the technical doc no longer holds true. The issue that prevented us from identifying the fix earlier was that [these tests](https://github.com/sablier-labs/flow/blob/81a6c016e94be218ba6038c281d2ca028c337298/tests/integration/concrete/withdraw-delay/withdrawDelay.t.sol#L10) were giving false positives due to descaling the ongoing debt back to the token's decimals.

For example, previously, if the withdrawable amount was 1.99 tokens, the recipient would withdraw 1 token, and the snapshot debt would remain 0, causing a delay for the remaining 0.99 tokens. Now, with the snapshot updated to 0.99, thus there is no delay which is proven by the new test in this branch: 

https://github.com/sablier-labs/flow/blob/bc6643b88114d9306052938149777ce882955d05/tests/integration/concrete/withdraw-delay/withdrawDelay.t.sol#L10-L47


tests/integration/fuzz/withdrawMultiple.t.sol
smol-ninja, line 1: 1. Can you also please update the checklist? It still mentions "less than".
2. What about updating `testFuzz_WithdrawMultiple_Usdc_SmallDelay` as well?

tests/integration/fuzz/withdrawMultiple.t.sol
```diff
@@ -82,7 +82,7 @@ contract WithdrawMultiple_Delay_Fuzz_Test is Shared_Integration_Fuzz_Test {
     )
         external
     {
-        _test_WithdrawMultiple(rps, withdrawCount, timeJump, decimals, ISablierFlow.withdrawMax.selector, 0);
+        _test_WithdrawMultiple(rps, withdrawCount, timeJump, decimals, 0);
```
smol-ninja, line 84: Why removed the selector? The goal was to test both.

tests/invariant/Flow.t.sol
```diff
@@ -333,7 +333,7 @@ contract Flow_Invariant_Test is Base_Test {
     }
 
     /// @dev For non-voided streams, the expected streamed amount should be greater than or equal to the sum of total
-    /// debt and withdrawn amount. And, the difference between the two should not exceed 10 mvt.
+    /// debt and withdrawn amount.
```
smol-ninja, line 336: It should be "equal" only here now.

tests/invariant/Flow.t.sol
smol-ninja, line 1: Could you also update the invariant section in the docs?

smol-ninja (changes requested):
Excellent observation @andreivladbrg. Left some comments below.

tests/integration/fuzz/withdrawMultiple.t.sol
andreivladbrg: no, the idea was to use equality in assertion, and if we keep the simple `withdraw` we can't achieve this

andreivladbrg: also, what is the point of using `withdraw`  if we want to test if there is delay?

smol-ninja: One reason why `withdraw` was included in this test was to make sure the actual streamed amount never exceed desired streamed amount i.e. no leaking of funds when rps is unrealistic. Previously, there were cases when `withdraw` was causing more delay than `withdrawMax` depending on with what amount and at what time you call them.

Therefore, even if there is no delay anymore, it may be a good idea to ensure that in case of `withdraw` the actual withdrawn is less than or equal to desired withdrawn. So this would solve the problem that we cant achieve equality in case of `withdraw`.

`withdraw` and `withdrawMax` can behave differently because one uses value of `coveredDebt` and another one takes input from user.

tests/integration/fuzz/withdrawMultiple.t.sol
```diff
@@ -50,60 +48,27 @@ contract WithdrawMultiple_Delay_Fuzz_Test is Shared_Integration_Fuzz_Test {
...
-        // Calculate the stream delay.
-        uint256 streamDelay = (deviationAmount * SCALE_FACTOR) / rps;
-
-        // Assert that the stream delay is within 5 second for the given fuzzed rps.
-        assertLe(streamDelay, 5 seconds);
-
-        // Assert that the deviation is less than 0.01e6 USDC.
-        assertLe(deviationAmount, 0.01e6);
-
         // Assert that actual withdrawn amount is always less than the desired amount.
```
smol-ninja, line 51: This comment requires update.

tests/invariant/Flow.t.sol
smol-ninja, line 1: [Technical doc](https://github.com/sablier-labs/flow/blob/test/delay-fix/TECHNICAL-DOC.md) needs to have updated invariant as well.

smol-ninja (changes requested):

tests/integration/fuzz/withdrawMultiple.t.sol
smol-ninja: On this @andreivladbrg, if you think that testing `withdrawMax` for errors also tests `withdraw` implicitly, then lets go with your changes. Otherwise, we should test for `withdraw` as well.

andreivladbrg: > it may be a good idea to ensure that in case of `withdraw` the actual withdrawn is less than or equal to desired with

well, ofc, but if the bounded amount is going to be 1, it for sure going to be lower than the lower bound :))
the point think you are missing, the param amount in the function cannot be greater than `coveredDebt` as we have other [concrete test](https://github.com/sablier-labs/flow/blob/004a94be20765ea8e5f90ba280bde510bc2cf972/tests/integration/concrete/withdraw/withdraw.t.sol#L96-L133) for it. and here, once again, we are testing if there is no delay (equality)

> anymore, it may be a good idea to ensure that in case of withdraw the actual withdrawn is less than or equal to desired withdrawn

we have a sum of withdrawn amounts. if the maximum possible scenario of this sum (only `withdrawMax` calls) equals the total amount streamed it implicitly test that the sum of bound amounts between `[1,max withdrawable]` (we were using `bound(withdrawAmount, 1, flow.withdrawableAmountOf)`) will always be lower or equal to total, thus introducing this test does not add value in what this new file should test "NoDelay" but it test other thing, i.e. withdraw param is lower than covered debt, which is also in this [invariant test](https://github.com/sablier-labs/flow/blob/dabed1f2bd219653bbe59a3f8813f2ad3daa396c/tests/invariant/Flow.t.sol#L337) and in the concrete tests mentioned above

lmk if it makes sense

smol-ninja: You convinced me on "param amount in the function cannot be greater than `coveredDebt`". Good point.

tests/integration/fuzz/withdrawMultiple.t.sol
```diff
@@ -50,60 +48,27 @@ contract WithdrawMultiple_Delay_Fuzz_Test is Shared_Integration_Fuzz_Test {
...
         uint128 rps,
-        uint128 withdrawAmount,
         uint256 withdrawCount,
         uint40 timeJump,
         uint8 decimals
     )
         external
     {
-        _test_WithdrawMultiple(rps, withdrawCount, timeJump, decimals, ISablierFlow.withdraw.selector, withdrawAmount);
+        _test_WithdrawMultiple(rps, withdrawCount, timeJump, decimals, 0);
```
smol-ninja, line 71: Lets remove the private function and add the logic directly to `testFuzz_WithdrawMaxMultiple_RpsWideRange_NoDelay`

smol-ninja (approved):
Good to go. I left a comment below in case you want to address it.

tests/integration/fuzz/withdrawMultiple.t.sol
andreivladbrg: good point, will update

Issue #299, opened 2024-10-13
Concrete tests for `SablierFlowBase` getters


smol-ninja:
Some of the getters in `SablierFlowBase` simply read and return storage values and so do not have their own concrete tests. As a consequence of which, we missed adding `notNull` to `isTransferable` function. This was discovered during [Cantina audit](https://cantina.xyz/code/99ae802b-f05c-4e36-a1d1-240d5146649c/findings/2).

To avoid similar errors in the future, we should concrete tests for those getters too.

PR #321, opened 2024-10-26
test: add getters tests

andreivladbrg:
Closes https://github.com/sablier-labs/flow/issues/299

### DIFF
--- a/tests/integration/concrete/getters/getters.t.sol
+++ b/tests/integration/concrete/getters/getters.t.sol
@@ -6,10 +6,6 @@ import { ud21x18 } from "@prb/math/src/UD21x18.sol";
 import { Integration_Test, Flow } from "../../Integration.t.sol";
 
 contract Getters_Integration_Concrete_Test is Integration_Test {
-    function setUp() public override {
-        Integration_Test.setUp();
-    }
-
     /*//////////////////////////////////////////////////////////////////////////
                                     GET-BALANCE
     //////////////////////////////////////////////////////////////////////////*/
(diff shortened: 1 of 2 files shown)


tests/integration/concrete/getters/getters.t.sol
smol-ninja, line 1: There are other getters as well. So how about being more explicit with the name here with `flowbase-getters/flowBaseGetters.t.sol`?

smol-ninja (changes requested):
Nice work on the PR. Very sorry I took so long to mention that bulloak supports multiple functions in the same test suite.

```tree
SablierFlowBase_Getters::getBalance
├── given null
│  └── it should revert
└── given not null
   ├── given zero
   │  └── it should return zero value
   └── given not zero
      └── it should return non-zero value


SablierFlowBase_Getters::isPaused
├── given null
│  └── it should revert
└── given not null
   ├── given true
   │  └── it should return true value
   └── given not true
      └── it should return false value
```

and then if you run 

```shell
bulloak scaffold -wf tests/integration/concrete/getters/getters.tree
```

it will generate a single contract to test all these getters. We wont have to make separate directories for each function without losing on BTT. I am in favour of keeping BTT for the same because it makes it easier to see whats tests and whats not. Let me know what you think.

tests/integration/concrete/getters/getters.t.sol
andreivladbrg: which one? are you saying that the function that calculate an amount (refundable/withdrawable) are also getter?
IMO i wouldn't call them getters only because they are _constant_ functions

andreivladbrg:
> bulloak supports multiple functions in the same test suite

ah lol, how did i not know?

> I am in favour of keeping BTT

now i’m in favor as well

tests/integration/concrete/getters/getters.t.sol
smol-ninja: You are right. They are not getters. lets keep it as `getters/getters.t.sol`

andreivladbrg:
@smol-ninja updated the branch with the `.tree` file

smol-ninja (approved):
Thank you for the changes. Lesss go 🚀

Issue #323, opened 2024-10-28
Implement `refundMax`


smol-ninja:
As discussed in https://github.com/sablier-labs/flow/discussions/322.

PR #324, opened 2024-10-28
feat: refundMax

smol-ninja:
Closes https://github.com/sablier-labs/flow/issues/323


andreivladbrg (approved):
you were so fast with this PR 

shogun_lgtm

PaulRBerg:
> you were so fast with this PR

I think Shub's GitHub handle has something to do with that

PR #326, opened 2024-11-02
added: IBatch Interface

### LINKED ISSUE #325: Implement `IBatch` interface
## To do

- [x] Implement `IBatch` interface
- [x] Inherit `IBatch` into `ISablierFlowBase`

TheMarvelFan:
Closes #325 

Added IBatch interface and inherited it to ISablierFlowBase inerface.
Also inherited IBatch into Batch to fix inheritance issues during compile-time.

### DIFF
--- a/src/abstracts/Batch.sol
+++ b/src/abstracts/Batch.sol
@@ -2,18 +2,20 @@
 pragma solidity >=0.8.22;
 
 import { Errors } from "../libraries/Errors.sol";
+import { IBatch } from "../interfaces/IBatch.sol";
 
 /// @title Batch
 /// @notice This contract implements logic to batch call any function.
 /// @dev Forked from: https://github.com/boringcrypto/BoringSolidity/blob/master/contracts/BoringBatchable.sol
-abstract contract Batch {
+abstract contract Batch is IBatch {
     /*//////////////////////////////////////////////////////////////////////////
                          USER-FACING NON-CONSTANT FUNCTIONS
     //////////////////////////////////////////////////////////////////////////*/
 
+    /// @inheritdoc IBatch
     /// @notice Allows batched call to self, `this` contract.
     /// @param calls An array of inputs for each call.
-    function batch(bytes[] calldata calls) external {
+    function batch(bytes[] calldata calls) external virtual override {
         uint256 count = calls.length;
 
         for (uint256 i = 0; i < count; ++i) {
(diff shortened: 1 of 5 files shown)


src/abstracts/Batch.sol
```diff
@@ -2,18 +2,20 @@
 pragma solidity >=0.8.22;
 
 import { Errors } from "../libraries/Errors.sol";
+import { IBatch } from "../interfaces/IBatch.sol";
 
 /// @title Batch
 /// @notice This contract implements logic to batch call any function.
```
andreivladbrg, line 8: we should say for the `notice` natpsec:

```solidity
/// @notice See the documentation in {IBatch}.
```

src/abstracts/Batch.sol
```diff
@@ -2,18 +2,20 @@
...
+abstract contract Batch is IBatch {
     /*//////////////////////////////////////////////////////////////////////////
                          USER-FACING NON-CONSTANT FUNCTIONS
     //////////////////////////////////////////////////////////////////////////*/
 
+    /// @inheritdoc IBatch
     /// @notice Allows batched call to self, `this` contract.
     /// @param calls An array of inputs for each call.
-    function batch(bytes[] calldata calls) external {
+    function batch(bytes[] calldata calls) external virtual override {
```
andreivladbrg, line 18: there is no need to use `virtual` here

src/interfaces/ISablierFlowBase.sol
```diff
@@ -18,7 +18,8 @@ import { IFlowNFTDescriptor } from "./IFlowNFTDescriptor.sol";
 interface ISablierFlowBase is
     IERC4906, // 2 inherited components
     IERC721Metadata, // 2 inherited components
-    IAdminable // 0 inherited components
+    IAdminable, // 0 inherited components
+    IBatch // 0 inherited components
```
andreivladbrg, line 22: it should be inherited in `ISablierFlow` instead, as the `Batch` contract is not inherited in `SablierFlowBase` but in `SablierFlow`

also, the inherited components count should be updated

andreivladbrg (changes requested):
Thank you for working on this, just left some comments bellow:

smol-ninja (approved):
LGTM. Thanks @TheMarvelFan for the PR. We use `slither` to fetch the number of inherited components.

```bash
slither src/SablierFlow.sol --print inheritance
```

andreivladbrg (approved):
👍

PR #348, opened 2024-12-13
feat: add eth fees in flow functions

### LINKED ISSUE #328: ETH fees in withdraw function
The design should be the same as in [Lockup](https://github.com/sablier-labs/v2-core/issues/1076).

### Discussed in https://github.com/sablier-labs/flow/discussions/327

<div type='discussions-op-text'>

<sup>Originally posted by **PaulRBerg** November  6, 2024</sup>
We've recently decided to charge [ETH fees](https://github.com/sablier-labs/v2-core/issues/1076) instead of [protocol fees](https://github.com/sablier-labs/v2-core/issues/1034) in Lockup. ETH fees make much more sense for Lockup streams per all the reasons provided [here](https://github.com/sablier-labs/company-discussions/discussions/72#discussioncomment-11115807).

Now, for Flow streams, the fee is paid in stablecoin. This is not bad, but I wonder if charging in ETH wouldn't be better?

1. More consistency across our fee models
2. Less psychological resistance; the user has to pay a fee in ETH anyway (for the gas), so charging $1 in ETH should feel less pressuring than paying a *separate* fee in USDC (from the recipient's salary)
3. Charging in ETH means that we are charging the account that withdraws, not the recipient (these accounts might not be the same)

Notes:

- I'm not suggesting to change the implementation now because doing so would cancel the validity of the security audits.
- We will start charging in ~March anyway, so changing the implementation in Flow now would be overkill.

Cc @sablier-labs/everybody for feedback.</div>

### Tasks
- [x] https://github.com/sablier-labs/flow/issues/338

andreivladbrg:
Closes #328 

**Note:** Added concrete tests for this to ensure that we don't miss marking a function as `payable`


PaulRBerg (changes requested):
Thanks Andrei.

Good idea to write concrete tests, but it's overkill to test these things:

- it should update the caller ETH balance
- it should update the flow contract ETH balance

Just having the call compiling with the `{ value } ` option would be enough. We are not meant to test that the EVM is implemented correctly.

So let's please simplify those tests.

tests/fork/Flow.t.sol
```diff
@@ -217,11 +224,15 @@ contract Flow_Fork_Test is Fork_Test {
...
         // Make sure the requirements are respected.
-        resetPrank({ msgSender: flow.getSender(streamId) });
+        address sender = flow.getSender(streamId);
+        resetPrank({ msgSender: sender });
         if (flow.isPaused(streamId)) {
             flow.restart(streamId, RATE_PER_SECOND);
         }
 
+        // Fund the sender to pay the fee.
+        vm.deal({ account: sender, newBalance: sender.balance + FEE });
```
smol-ninja, line 234: Can we deal enough ETH to `sender` in Base contract to avoid dealing in each test?

smol-ninja (changes requested):
This PR should be created against a new branch called `staging`.

tests/fork/Fork.t.sol
```diff
@@ -91,6 +93,7 @@ abstract contract Fork_Test is Base_Test {
         address sender = flow.getSender(streamId);
         resetPrank({ msgSender: sender });
         deal({ token: address(token), to: sender, give: depositAmount });
+        vm.deal({ account: sender, newBalance: sender.balance + FEE });
```
smol-ninja, line 96: Same as https://github.com/sablier-labs/flow/pull/348#discussion_r1885083619. If we can deal enough ETH to sender through Base, we can remove this line.

tests/integration/concrete/collect-fees/collectFees.tree
```diff
@@ -0,0 +1,12 @@
+CollectFees_Integration_Concrete_Test
+├── when admin is not contract
```
smol-ninja, line 2: should be `given` because `admin` is a contract state and not a user input.

tests/integration/concrete/collect-fees/collectFees.t.sol
```diff
@@ -0,0 +1,74 @@
...
+            )
+        );
+
+        // Collect the fees.
+        flow.collectFees();
+    }
+
+    function test_WhenAdminImplementsReceiveFunction() external whenAdminIsContract {
+        // Transfer the admin to a contract that implements the receive function.
+        resetPrank({ msgSender: users.admin });
```
smol-ninja, line 40: Since this is common across all tests here, lets move this to `setUp`.

tests/integration/concrete/collect-fees/collectFees.t.sol
```diff
@@ -0,0 +1,74 @@
...
+        vm.warp({ newTimestamp: WITHDRAW_TIME });
+
+        // Load the initial ETH balance of the admin.
+        uint256 initialAdminBalance = admin.balance;
+
+        // Make recipient the caller.
+        resetPrank({ msgSender: users.recipient });
+
+        // Make a withdrawal and pay the fee.
+        flow.withdrawMax{ value: FEE }({ streamId: defaultStreamId, to: users.recipient });
```
smol-ninja, line 60: How about moving this to set up function? Withdraw leads sends fee to the contract and then collect fee is called by the admin. So it should be a part of the setup which sets up the testing scenario and then in test functions, we just test `collectFees`.

tests/integration/concrete/payable/payable.t.sol
```diff
@@ -0,0 +1,112 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity >=0.8.22 <0.9.0;
+
+import { ud21x18 } from "@prb/math/src/UD21x18.sol";
+
+import { Shared_Integration_Concrete_Test } from "../Concrete.t.sol";
+
+contract Payable_Integration_Concrete_Test is Shared_Integration_Concrete_Test {
```
smol-ninja, line 8: The fork test that you have modified also cover this particular scenario. If any of the functions misses `payable`, the fork tests would fail. So imo we can remove it.

smol-ninja (changes requested):
Sorry, I accidentally clicked submit review. Here is full review.

tests/integration/concrete/payable/payable.t.sol
andreivladbrg: The fork tests do not cover all 15 unique functions

tests/integration/concrete/collect-fees/collectFees.tree
andreivladbrg: Should be updated in lockup as well, then

tests/fork/Flow.t.sol
andreivladbrg: The sender is fuzzed in create, we don't have an unique sender

tests/integration/concrete/payable/payable.t.sol
smol-ninja: You are right.

tests/integration/concrete/payable/payable.t.sol
```diff
@@ -0,0 +1,112 @@
...
+    function _test_ETHBalance(bytes memory callData) private {
+        // Load the initial ETH balance of the sender.
+        uint256 initialSenderBalance = users.sender.balance;
+        uint256 initialFlowBalance = address(flow).balance;
+
+        (bool success,) = address(flow).call{ value: FEE }(callData);
+        assertTrue(success, "payable call failed");
+
+        // Assert that both the sender and the contract have the expected balances.
+        assertEq(users.sender.balance, initialSenderBalance - FEE);
```
smol-ninja, line 28: Based on https://github.com/sablier-labs/flow/pull/348#discussion_r1885087102, this test is good for me. But should we remove assertions? If a function does not have `payable`, `success` would not be `true`. So no need to check for balances.

tests/integration/concrete/collect-fees/collectFees.tree
smol-ninja: Yes. But we can do that later in lockup.

tests/fork/Flow.t.sol
smol-ninja: Can we fund the fuzzed sender right after it is created? After `checkUsers(params.recipient, params.sender);` line?

andreivladbrg: yeah, we could, correct

andreivladbrg:
> Good idea to write concrete tests, but it's overkill to test these things:
it should update the caller ETH balance
it should update the flow contract ETH balance
Just having the call compiling with the { value }  option would be enough. We are not meant to test that the EVM is implemented correctly.

thanks. fair, let's remove them. also, does it mean we should remove the `address.balance` check from all places?

andreivladbrg:
@PaulRBerg @smol-ninja updated the branch with your feedback, lmk if it looks good

smol-ninja (approved):
LGTM.

Issue #354, opened 2024-12-21
Important adjustments to the batch function


PaulRBerg:
Same as https://github.com/sablier-labs/v2-core/issues/1124.

smol-ninja:
@PaulRBerg just to confirm, this issue is a part of Flow 1.2.0 (Mar/Apr release) and not Flow 1.1.0 (Jan release) whereas https://github.com/sablier-labs/v2-core/issues/1124 is a part of Lockup 2.0.0 (Jan release) or is that also a part of Lockup 2.1.0 (future release)?

PaulRBerg:
No, I meant to include this as part of Flow v1.1.0

All the problems mentioned in https://github.com/sablier-labs/v2-core/issues/1124 apply to Flow, too.

smol-ninja:
Sounds good. I will work on it after https://github.com/sablier-labs/v2-core/pull/1126 is merged.

PR #358, opened 2024-12-31
refactor: batch function

smol-ninja:
Closes https://github.com/sablier-labs/flow/issues/354.

Similar to `Adminable` in Flow, unit tests for `Batch` are skipped. In future, common contracts across the Sablier protocols will be moved to a separate [Utilities repo](https://github.com/orgs/sablier-labs/discussions/13) which will fix this.

### DIFF
--- a/src/abstracts/Batch.sol
+++ b/src/abstracts/Batch.sol
@@ -1,26 +1,39 @@
 // SPDX-License-Identifier: GPL-3.0-or-later
+// solhint-disable no-inline-assembly
 pragma solidity >=0.8.22;
 
 import { IBatch } from "../interfaces/IBatch.sol";
-import { Errors } from "../libraries/Errors.sol";
 
 /// @title Batch
 /// @notice See the documentation in {IBatch}.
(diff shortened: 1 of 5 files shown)


andreivladbrg:
> Instead of rewriting batch contract, I decided to import it from https://github.com/sablier-labs/v2-core/pull/1126 to avoid rewriting all the unit tests here as well

Hmm, I have mixed feelings about this. I wouldn’t add the `lockup` contracts as a dep just for this. They’re different systems, and adding it could create confusion. Plus, we’re going to start using the utils repo soon anyway.

smol-ninja:
I also had mixed feelings about it so I completely agree with you. I will update the PR once https://github.com/sablier-labs/v2-core/pull/1126 is merged.

smol-ninja:
This PR is ready @andreivladbrg, so feel free to review it now

src/abstracts/Batch.sol
PaulRBerg, line 1: I used DiffChecker to check that the source code of this file is identical to the one in Lockup, and it was.

PaulRBerg (approved):
looking good

Issue #339, opened 2024-11-25
Include a note on reliability of `statusOf` function


smol-ninja:
The return value of `statusOf` cannot be relied upon. It can be manipulated by a smart sender. Take the following example:

1. A call to `statusOf` is made inside a transaction.
2. The sender sees the pending tx and front runs it by pausing the stream.
3. The function returns `PAUSED`.
4. The sender sends another tx in the next block to resume the stream.

Therefore, the natspecs of `statusOf` should mention a note about it as the following:

```solidity
/// Since Flow offers flexibility to the senders, a sender can pause and resume a stream at any time. Therefore, 
/// it is not recommended to rely solely on the return value of `statusOf`.
```

PR #363, opened 2025-01-07
docs: include a warning note for statusOf

smol-ninja:
Closes https://github.com/sablier-labs/flow/issues/339.

Can we include this in the upcoming release? Its a small note about `statusOf`.


PaulRBerg (approved):
yeah, let's do it

Issue #238, opened 2024-09-13
Polish withdraw tree


smol-ninja:
Related:
- https://github.com/sablier-labs/flow/pull/232#discussion_r1756552059
- https://github.com/sablier-labs/flow/pull/232#discussion_r1756553109
- https://github.com/sablier-labs/flow/pull/234#discussion_r1773046717

smol-ninja:
```tree
            └── when withdrawal address owner
               ├── when amount overdraws
               │  ├── given stream has uncovered debt
               │  │  └── it should revert
               │  └── given stream has no uncovered debt
               │     └── it should revert
               └── when amount not overdraw
                  ├── when amount not equal total debt
                  │  ├── it should update snapshot debt
                  │  └── it should make the withdrawal
                  └── when amount equal total debt
                     ├── when amount not exceed snapshot debt
                     │   ├── it should not update snapshot time
                     │   └── it should make the withdrawal
                     └── when amount exceed snapshot debt
                        ├── given protocol fee not zero
                        │  ├── it should update the protocol revenue
                        │  └── it should withdraw the net amount
                        └── given protocol fee zero
                           ├── given token has 18 decimals
                           │  └── it should make the withdrawal
                           └── given token not have 18 decimals
                              ├── it should update snapshot debt to zero
                              ├── it should update snapshot time
```

vs 

```tree
            └── when withdrawal address owner
               ├── given balance not exceed total debt
               │  ├── when amount exceeds balance
               │  │  └── it should revert
               │  └── when amount not exceed balance
               │     └── it should withdraw
               └── given balance exceeds total debt
                  ├── when amount greater than total debt
                  │  └── it should revert
                  ├── when amount equals total debt
                  │  ├── it should make the withdrawal
                  │  ├── it should update snapshot debt to zero
                  │  └── it should update snapshot time to current time
                  └── when amount less than total debt
                     ├── when amount less than snapshot debt
                     │  ├── it should make the withdrawal
                     │  ├── it should not update snapshot time
                     │  └── it should reduce snapshot debt by amount withdrawn
                     ├── when amount equals snapshot debt
                     │  ├── it should make the withdrawal
                     │  ├── it should not update snapshot time
                     │  └── it should update snapshot debt to zero
                     └── when amount greater than snapshot debt
                        ├── given protocol fee not zero
                        │  ├── it should update the protocol revenue
                        │  └── it should withdraw the net amount
                        └── given protocol fee zero
                           ├── given token has 18 decimals
[...]

PR #368, opened 2025-01-10
test: refactor withdraw tests to cover all edge cases

smol-ninja:
### Changelog
- Closes https://github.com/sablier-labs/flow/issues/238.
- Renames `updateSnapshotTimeAndWarp` function to `takeSnapshot` and removes `warp` logic from it.
- Renames `WARP_ONE_MONTH` constant to `ONE_MONTH_SINCE_START`. 
    - The variable name `WARP_ONE_MONTH` goes well when called in `vm.warp` but not when used as a variable. For example in `uint40 exptectedDepletionTime = WARP_ONE_MONTH + uint40(solvencyPeriod + 1)`.

### DIFF
--- a/tests/integration/Integration.t.sol
+++ b/tests/integration/Integration.t.sol
@@ -62,18 +62,15 @@ abstract contract Integration_Test is Base_Test {
         deposit(streamId, depositAmount);
     }
 
-    /// @dev Update the snapshot using `adjustRatePerSecond` and then warp block timestamp to it.
-    function updateSnapshotTimeAndWarp(uint256 streamId) internal {
+    /// @dev Take the snapshot using `adjustRatePerSecond`.
+    function takeSnapshot(uint256 streamId) internal {
         resetPrank(users.sender);
         UD21x18 ratePerSecond = flow.getRatePerSecond(streamId);
 
-        // Updates the snapshot time via `adjustRatePerSecond`.
+        // Take the snapshot via `adjustRatePerSecond`.
         flow.adjustRatePerSecond(streamId, ud21x18(1));
 
         // Restores the rate per second.
         flow.adjustRatePerSecond(streamId, ratePerSecond);
-
-        // Warp to the snapshot time.
-        vm.warp({ newTimestamp: flow.getSnapshotTime(streamId) });
     }
 }
--- a/tests/integration/Integration.t.sol
+++ b/tests/integration/Integration.t.sol
@@ -64,6 +64,10 @@ abstract contract Integration_Test is Base_Test {
 
     /// @dev Updates the snapshot time and snapshot debt by temporarily adjusting the rate per second..
     function updateSnapshot(uint256 streamId) internal {
+        // Read the current caller.
+        (, address originalCaller,) = vm.readCallers();
+
+        // Switch to the sender and adjust the rate per second.
         resetPrank(users.sender);
         UD21x18 ratePerSecond = flow.getRatePerSecond(streamId);
 
@@ -72,5 +76,8 @@ abstract contract Integration_Test is Base_Test {
 
         // Restore the original rate per second.
         flow.adjustRatePerSecond(streamId, ratePerSecond);
+
+        // Switch back to the original caller.
+        resetPrank(originalCaller);
     }
 }
--- a/tests/integration/Integration.t.sol
+++ b/tests/integration/Integration.t.sol
@@ -62,15 +62,15 @@ abstract contract Integration_Test is Base_Test {
         deposit(streamId, depositAmount);
     }
 
-    /// @dev Take the snapshot using `adjustRatePerSecond`.
-    function takeSnapshot(uint256 streamId) internal {
+    /// @dev Updates the snapshot time and snapshot debt by temporarily adjusting the rate per second..
+    function updateSnapshot(uint256 streamId) internal {
         resetPrank(users.sender);
         UD21x18 ratePerSecond = flow.getRatePerSecond(streamId);
 
-        // Take the snapshot via `adjustRatePerSecond`.
+        // Take the snapshot by temporarily setting the rate per second to 1.
         flow.adjustRatePerSecond(streamId, ud21x18(1));
 
-        // Restores the rate per second.
+        // Restore the original rate per second.
         flow.adjustRatePerSecond(streamId, ratePerSecond);
     }
 }
--- a/tests/integration/concrete/withdraw/withdraw.t.sol
+++ b/tests/integration/concrete/withdraw/withdraw.t.sol
@@ -3,6 +3,7 @@ pragma solidity >=0.8.22;
 
 import { IERC4906 } from "@openzeppelin/contracts/interfaces/IERC4906.sol";
 import { IERC20 } from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+import { ud, ZERO } from "@prb/math/src/UD60x18.sol";
 
 import { ISablierFlow } from "src/interfaces/ISablierFlow.sol";
 import { Errors } from "src/libraries/Errors.sol";
@@ -13,28 +14,39 @@ contract Withdraw_Integration_Concrete_Test is Shared_Integration_Concrete_Test 
     function setUp() public override {
         Shared_Integration_Concrete_Test.setUp();
 
-        depositToDefaultStream();
+        // Deposit amount equals to one months of streaming.
+        deposit(defaultStreamId, ONE_MONTH_DEBT_6D);
+
+        // Take a snapshot after one month of streaming.
+        takeSnapshot(defaultStreamId);
+
+        // Forward time by one more month, so that total debt becomes (2 * ONE_MONTH_DEBT_6D).
+        vm.warp({ newTimestamp: getBlockTimestamp() + ONE_MONTH });
 
         // Set recipient as the caller for this test.
         resetPrank({ msgSender: users.recipient });
     }
 
     function test_RevertWhen_DelegateCall() external {
-        bytes memory callData = abi.encodeCall(flow.withdraw, (defaultStreamId, users.recipient, WITHDRAW_TIME));
+        // It should revert.
+        bytes memory callData = abi.encodeCall(flow.withdraw, (defaultStreamId, users.recipient, WITHDRAW_AMOUNT_6D));
         expectRevert_DelegateCall(callData);
     }
 
     function test_RevertGiven_Null() external whenNoDelegateCall {
-        bytes memory callData = abi.encodeCall(flow.withdraw, (nullStreamId, users.recipient, WITHDRAW_TIME));
+        // It should revert.
+        bytes memory callData = abi.encodeCall(flow.withdraw, (nullStreamId, users.recipient, WITHDRAW_AMOUNT_6D));
         expectRevert_Null(callData);
     }
 
     function test_RevertWhen_AmountZero() external whenNoDelegateCall givenNotNull {
+        // It should revert.
         vm.expectRevert(abi.encodeWithSelector(Errors.SablierFlow_WithdrawAmountZero.selector, defaultStreamId));
         flow.withdraw({ streamId: defaultStreamId, to: users.recipient, amount: 0 });
     }
 
     function test_RevertWhen_WithdrawalAddressZero() external whenNoDelegateCall givenNotNull whenAmountNotZero {
+        // It should revert.
         vm.expectRevert(abi.encodeWithSelector(Errors.SablierFlow_WithdrawToZeroAddress.selector, defaultStreamId));
         flow.withdraw({ streamId: defaultStreamId, to: address(0), amount: WITHDRAW_AMOUNT_6D });
     }
@@ -49,6 +61,7 @@ contract Withdraw_Integration_Concrete_Test is Shared_Integration_Concrete_Test 
     {
         resetPrank({ msgSender: users.sender });
 
+        // It should revert.
         vm.expectRevert(
             abi.encodeWithSelector(
                 Errors.SablierFlow_WithdrawalAddressNotRecipient.selector, defaultStreamId, users.sender, users.sender
@@ -67,6 +80,7 @@ contract Withdraw_Integration_Concrete_Test is Shared_Integration_Concrete_Test 
     {
         resetPrank({ msgSender: users.eve });
 
+        // It should revert.
         vm.expectRevert(
             abi.encodeWithSelector(
                 Errors.SablierFlow_WithdrawalAddressNotRecipient.selector, defaultStreamId, users.eve, users.eve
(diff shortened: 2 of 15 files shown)


tests/integration/concrete/withdraw/withdraw.tree
```diff
@@ -14,36 +14,45 @@ Withdraw_Integration_Concrete_Test
             ├── when withdrawal address not owner
             │  ├── when caller sender
             │  │  └── it should revert
-            │  ├── when caller unknown
-            │  │  └── it should revert
-            │  └── when caller recipient
-            │     └── it should withdraw
```
andreivladbrg, line 20: why was this branch removed from here?

there is no test for the case when:

`msg.sender == recipient && to != recipient`

if it is due to bulloak, i think we can use the "caller owner" terminology

tests/utils/Vars.sol
```diff
@@ -11,6 +11,7 @@ struct Vars {
     // previous values.
     uint256 previousAggregateAmount;
     uint256 previousOngoingDebtScaled;
+    uint256 previousSnapshotDebtScaled;
```
andreivladbrg, line 14: why was this added? 
it is not used anywhere

tests/integration/concrete/withdraw/withdraw.t.sol
```diff
@@ -186,32 +256,37 @@ contract Withdraw_Integration_Concrete_Test is Shared_Integration_Concrete_Test
...
+        whenCallerRecipient
+        givenBalanceExceedsTotalDebt
+        whenAmountLessThanTotalDebt
+        whenAmountGreaterThanSnapshotDebt
     {
-        // Go back to the starting point.
-        vm.warp({ newTimestamp: OCT_1_2024 });
-
-        resetPrank({ msgSender: users.sender });
+        // Turn on the protocol fee for usdc.
```
andreivladbrg, line 264: should we capitalize to `USDC`?

tests/integration/concrete/withdraw/withdraw.t.sol
```diff
@@ -251,87 +333,88 @@ contract Withdraw_Integration_Concrete_Test is Shared_Integration_Concrete_Test
...
+        vars.expectedProtocolRevenue = flow.protocolRevenue(vars.token) + vars.expectedProtocolFeeAmount;
+        vars.expectedWithdrawAmount = withdrawAmount - vars.expectedProtocolFeeAmount;
 
         // It should emit 1 {Transfer}, 1 {WithdrawFromFlowStream} and 1 {MetadataUpdated} events.
         vm.expectEmit({ emitter: address(vars.token) });
-        emit IERC20.Transfer({ from: address(flow), to: to, value: withdrawAmount - protocolFeeAmount });
+        emit IERC20.Transfer({ from: address(flow), to: to, value: vars.expectedWithdrawAmount });
 
-        vm.expectEmit({ emitter: address(flow) });
+        // vm.expectEmit({ emitter: address(flow) });
```
andreivladbrg: i think typo
same below

tests/integration/concrete/withdraw/withdraw.t.sol
```diff
@@ -251,87 +333,88 @@ contract Withdraw_Integration_Concrete_Test is Shared_Integration_Concrete_Test
...
-            protocolFeeAmount: 0,
-            withdrawAmount: WITHDRAW_AMOUNT_6D
+            withdrawAmount: amountToWithdraw
         });
+
+        // It should set snapshot debt to difference between total debt and amount withdrawn.
+        assertEq(
+            flow.getSnapshotDebtScaled(defaultStreamId),
+            getScaledAmount(totalDebt - amountToWithdraw, 6),
+            "snapshot debt"
```
andreivladbrg, line 357: why not move this assert in `_test_Withdraw` since, it is suppose to be in 18 decimals

and keep only the `snapshot time` assert to individual tests

tests/integration/concrete/withdraw/withdraw.t.sol
```diff
@@ -251,87 +333,88 @@ contract Withdraw_Integration_Concrete_Test is Shared_Integration_Concrete_Test
...
+        );
+        // It should update snapshot time to current time
+        assertEq(flow.getSnapshotTime(defaultStreamId), getBlockTimestamp(), "snapshot debt");
```
andreivladbrg: incorrect error string reason

should be "snapshot time" - similar across the file

andreivladbrg (commented):
great work, left some comments bellow:

tests/integration/concrete/withdraw/withdraw.t.sol
andreivladbrg: bf259891f327b49acc808599ba0be915e54e50eb

tests/integration/concrete/withdraw/withdraw.t.sol
andreivladbrg: bf259891f327b49acc808599ba0be915e54e50eb

tests/utils/Vars.sol
smol-ninja: Good catch. I added it to use but ended up not using it.

tests/integration/concrete/withdraw/withdraw.t.sol
smol-ninja: There are three places: `test_WhenAmountEqualsTotalDebt`, `test_WhenAmountLessThanSnapshotDebt` and `test_WhenAmountEqualsSnapshotDebt` where there are different assertions for snapshot debt. Thats why I put them into individual tests.

tests/integration/concrete/withdraw/withdraw.tree
smol-ninja: I made the following changes:

1. When withdrawal address is not owner: check for sender, unknown and recipient callers because it should revert for the first two.
2. When withdrawal address is owner: use only recipient as the caller. Caller is no more relevant in this case. For completeness, we have `testFuzz_Withdraw` where we are fuzzing caller as well.
    - I could run tests for all three callers using a modifier but then it makes the code ugly and difficult to read since I will have to go back and forth in time to reset stream IDs for each caller.

tests/integration/Integration.t.sol
```diff
@@ -62,18 +62,15 @@ abstract contract Integration_Test is Base_Test {
         deposit(streamId, depositAmount);
     }
 
-    /// @dev Update the snapshot using `adjustRatePerSecond` and then warp block timestamp to it.
-    function updateSnapshotTimeAndWarp(uint256 streamId) internal {
+    /// @dev Take the snapshot using `adjustRatePerSecond`.
```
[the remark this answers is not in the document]
smol-ninja: https://github.com/sablier-labs/flow/pull/368/commits/f20dd8434393aa691c8a1663d09b1b9c318e022b: Also renamed function name to `updateSnapshot`.

tests/integration/concrete/withdraw/withdraw.t.sol
smol-ninja: https://github.com/sablier-labs/flow/pull/368/commits/f20dd8434393aa691c8a1663d09b1b9c318e022b

tests/integration/concrete/withdraw/withdraw.tree
andreivladbrg: > Caller is no more relevant in this case. 
hmm, i am not sure why say this i believe we should test each individual case here as well

> For completeness, we have testFuzz_Withdraw where we are fuzzing caller as well

IMO fuzz tests are meant to catch cases we couldn't think of in unit or integration tests

andreivladbrg: btw, if we keep the current version we could remove the `to` param in the `_test_Withdraw` function as `users.recipient` is passed on each test

tests/integration/concrete/withdraw/withdraw.t.sol
```diff
@@ -232,16 +299,20 @@ contract Withdraw_Integration_Concrete_Test is Shared_Integration_Concrete_Test
...
-            protocolFeeAmount: 0,
-            withdrawAmount: WITHDRAW_AMOUNT_18D
-        });
+            withdrawAmount: WITHDRAW_AMOUNT_18D // < total debt and > snapshot debt
+         });
+
+        // It should set snapshot debt to difference between total debt and amount withdrawn.
+        assertEq(flow.getSnapshotDebtScaled(streamId), ONE_MONTH_DEBT_18D - WITHDRAW_AMOUNT_18D, "snapshot debt");
+        // It should update snapshot time to current time
+        assertEq(flow.getSnapshotTime(defaultStreamId), getBlockTimestamp(), "snapshot time");
```
andreivladbrg, line 315: here we are testing the same stream:

```solidity
//                                     ↓ notice the ID
assertEq(flow.getSnapshotDebtScaled(streamId)
//                                     ↓ notice the ID
assertEq(flow.getSnapshotTime(defaultStreamId)
```

andreivladbrg (commented):
just noticed one more thing

tests/integration/concrete/withdraw/withdraw.t.sol
andreivladbrg: you are correct, we should keep this way

tests/integration/concrete/withdraw/withdraw.tree
smol-ninja: > why say this i believe we should test each individual case here as well

I think, in concrete tests, whether a case should be tested or not should depend on if it directly influences the outcome or not. So when withdrawal address is not owner, value of caller impacts the result whereas, when its not owner, value of caller does not impact the result. 

Consider a simple example. Suppose there’s a deposit function, `f(amount)`, that accepts deposits of any amount including 0. There, there is another function, `g(amount)`, which accepts deposits of any amount but not 0. 

Now in the case of `f(amount)`, will we be testing how the function behaves when amount = 0, in concrete tests? I’d say probably not (we don't check for all values for which tests work). However, for `g(amount)` we would absolutely check if it reverts when amount = 0. Wdyt? 

PS: These are just some thoughts I’m brainstorming so feel free to share any opinions or feedback you have.

---------------- 

> if we keep the current version we could remove the to param in the _test_Withdraw function as users.recipient is passed on each test

`test_WhenCallerRecipient` calls `_test_Withdraw` with `users.eve` as the recipient.

tests/integration/concrete/withdraw/withdraw.t.sol
smol-ninja: yes it should be `streamId`

smol-ninja: https://github.com/sablier-labs/flow/pull/368/commits/3f9bac38ddcd3e9c7e6ee3cf6cd74d10a85df7a9

tests/integration/concrete/withdraw/withdraw.tree
andreivladbrg: > Now in the case of f(amount), will we be testing how the function behaves when amount = 0, in concrete tests?

i don't think it is the same. a better comparison would have been:
1. `g(amount)` does not accept `amount == 0`
2. `f(amount)` accepts `amount == 0` only a certain criteria is met
   - and we test the case when that criteria is met

*(similar to our example)*
which IMO we should test.

---

> test_WhenCallerRecipient calls _test_Withdraw with users.eve as the recipient

right, missed it

smol-ninja: OK I an convinced. I will run these tests for all three.

andreivladbrg: thank you

smol-ninja: https://github.com/sablier-labs/flow/pull/368/commits/002a055d11fafc1acd171fa7a911121ccc5739fb

smol-ninja:
@andreivladbrg I have pushed a new commit, https://github.com/sablier-labs/flow/pull/368/commits/002a055d11fafc1acd171fa7a911121ccc5739fb, that resolved your comment about running test for all users as well as few minor changes.

Please let me know if it looks good now.

PS: I have added assertions to make sure withdraw amount satisfies conditions for each test.

smol-ninja:
Thanks.

1. Why did you replace eve with operator where the caller is unknown? Isn't eve more appropriate to represent an unknown caller?
2. Those assert statements were left intentional to confirm the test constraints. The assert that you added in `_test_Withdraw` isn't much useful because that will always be true.

andreivladbrg:
> Why did you replace eve with operator where the caller is unknown? Isn't eve more appropriate to represent an unknown caller?
 
as the commit message says, we use `users.eve` when a test should revert (eve from "evil" 😈)

> Those assert statements were left intentionally to verify the test constraints. The assert that you added in _test_Withdraw isn't much useful because it will always be true. 

why would using `assert` be useful and `assertTrue` not ? they do the same thing

> So I think we should bring them back as they verify the test constraints

`assertTrue` can also verify the tests constraints, but in case it reverts, foundry's version is friendlier with debugging

smol-ninja:
> as the commit message says, we use `users.eve` when a test should revert (eve from "evil" 😈)

Ooh. I thought eve is cute 🤣

> why would using assert be useful and assertTrue not ? they do the same thing

They are same but not the conditions. In https://github.com/sablier-labs/flow/pull/368/commits/247f716f26b5788f53a7987a05edf621eb224eef, some of the asserts check for strict inequality whereas the one you added checks for `<=`. I didn't have objection with using `assertTrue` over `assert` but the conditions that it is checking.

andreivladbrg:
> I didn't have objection with using assertTrue over assert but the conditions that it is checking

ok, i get it now, this is what i misunderstood. i thought you were talking specifically about `assert` vs `assertTrue`

to not have repetitive code what if we add boolean (`waEqTd` - withdraw amount equal total debt) to the function param, and then:

```solidity
if (waEqTd) {
    assertTrue(withdrawAmount <= flow.totalDebtOf(streamId), "amount <= total debt");
} else {
    assertTrue(withdrawAmount < flow.totalDebtOf(streamId), "amount < total debt");
}
```

wdyt?

smol-ninja:
You will still end up with more lines of codes. We have one assertion per test, we will now end up having one bool update per test plus 5 additional lines in `_test_Withdraw`.

I think having assertions in individual test is a better approach.

- We simulate time and set things up for `_withdraw`
- We test if `withdrawAmount` follows constraints
- If yes, then we call `_withdraw`.

And not all tests have constraints related to total debt. For example, `test_WhenAmountLessThanSnapshotDebt` checks if `amount < initialSnapshotDebt`.

andreivladbrg:
TBH, i don’t fully get why we need these test constraints verifications in the first place since (i) we are in control of the vars passed as `wa`, and (ii) we have the snapshot debt assertions after calling `_test_Withdraw`.
could you help be with this clarification first, please.

another thing is the use of flow functions for params before calling `flow.withdraw`. IMO, this isn’t the best approach for _concrete/unit tests_, where we should have exact values (defaults) for before/after

so, instead of these:

```solidity
uint256 initialSnapshotDebt = getDescaledAmount(flow.getSnapshotDebtScaled(defaultStreamId), 6);
uint128 withdrawAmount = uint128(initialSnapshotDebt) + WITHDRAW_AMOUNT_6D;
// to:
uint256 withdrawAmount = 3092e6;
```

smol-ninja:
>  i don’t fully get why we need these test constraints verifications in the first place since

I'd say we don't need it as such because now tests are complete so we can remove them for sure. But given that each test is run for three users so having them while writing tests helped me with debugging to verify if all constraints are matched. 

Using these assertions also led me to make changes [like these](https://github.com/sablier-labs/flow/blob/247f716f26b5788f53a7987a05edf621eb224eef/tests/integration/concrete/withdraw/withdraw.t.sol#L113), where we warp time after taking snapshot to ensure there is ongoing debt as well. So, I don't have problems removing them but they will be helpful in the future if we make changes to the tests and want to be sure of the constraints.

> we should have exact values (defaults) for before/after

Whats wrong with it? Why is it not the best approach? Also, this is why assertions are useful here because its hard to keep track of all the variables in absolute numbers with three users.

Can I also know why you don't want to keep assertions?

andreivladbrg:
good point about modifiers, will move back the assert in the individual functions

andreivladbrg (approved):
👍

smol-ninja:
Great. Thank you.

Issue #366, opened 2025-01-09
Remove the fees in the stream's token (the protocol fees)


andreivladbrg:
<div type='discussions-op-text'>

<sup> Originally discussed in https://github.com/sablier-labs/flow/discussions/365 </sup>

Remove the fees in the stream's token, historically referred to as 'protocol fees'.

PR #384, opened 2025-02-05
refactor: remove broker fee functionality

andreivladbrg:
Closes https://github.com/sablier-labs/flow/issues/375

PR #385, opened 2025-02-05
refactor: remove protocol fee

andreivladbrg:
Closes https://github.com/sablier-labs/flow/issues/366

Depends on #384


tests/utils/Constants.sol
```diff
@@ -2,25 +2,21 @@
...
-import { UD60x18, ud } from "@prb/math/src/UD60x18.sol";
 
 abstract contract Constants {
     // Amounts
     uint128 internal constant DEPOSIT_AMOUNT_18D = 50_000e18;
     uint128 internal constant DEPOSIT_AMOUNT_6D = 50_000e6;
-    uint256 internal constant FEE = 0.001e18;
+
+    // Fee
+    uint256 internal constant FEE = 0.0001e18;
```
smol-ninja: Why'd you change the fee?

andreivladbrg: by mistake i guess

andreivladbrg:
your commit will cause conflicts in the next PR 🙈

smol-ninja:
> your commit will cause conflicts in the next PR

Oh sorry. But I guess it will be very small one as I only changes one line. 😅 Feel free to revert it back.

tests/utils/Constants.sol
andreivladbrg: since it is going to be remove in the upcoming `build/install-evm-utils` i would simply leave it like this

PaulRBerg:
For historical reference and clarity — the protocol fees removed in this PR refer to the old fees charged in the underlying token (the token being streamed), NOT in the native token (e.g. ETH).

Issue #361, opened 2025-01-06
Add an `ERC20.transferFrom` wrapper


smol-ninja:
<sup>Originally discussed in https://github.com/sablier-labs/flow/discussions/360</sup>

Include a token transfer function (an `ERC20.transferFrom` wrapper) in the contract.

```solidity
function transferFrom(IERC20 token, address to, uint256 value) external payable {
  token.transferFrom(msg.sender, to, value);
}
```

smol-ninja:
@razgraf regarding this issue, 

- We cannot implement a wrapper for `ERC20.transfer` because from this context, the `msg.sender` will be Flow contract.
- However, we can implement `ERC20.transferFrom` but that would require user to first approve Flow contract to spend the token. Is that OK? In that case, should we call just it `transferFrom`?

```solidity
function transferFrom(IERC20 token, address to, uint256 value) external {
  token.transferFrom(msg.sender, to, value);
}
```

razgraf:
That's still great! In the context of paying out bonuses or one-off reimbursements, being able to track them and assign to a sender-recipient relationship become very powerful.

So I'm ok with the allowance being required. 

Thinking ahead but... can we also make it `payable` :D?

smol-ninja:
We will have to make it `payable` regardless because it can be batched with other functions.

PaulRBerg:
The things we have to do because EOAs cannot batch txs

smol-ninja:
@razgraf do you want to call it something else like `payTo` or `payDirectly` or `transferBonus` or anything else? I have called it `transferFrom` but I think something bespoke would be nice too.

PR #391, opened 2025-02-11
feat: include token transfer wrapper

smol-ninja:
Closes https://github.com/sablier-labs/flow/issues/361

### DIFF
--- a/src/SablierFlow.sol
+++ b/src/SablierFlow.sol
@@ -405,6 +405,12 @@ contract SablierFlow is
     }
 
     /// @inheritdoc ISablierFlow
+    function transferFrom(IERC20 token, address to, uint128 amount) external payable {
+        // Interaction: transfer the amount using ERC20's `transferFrom` function.
+        token.transferFrom(msg.sender, to, amount);
+    }
+
+    /// @inheritdoc ISablierFlow
     function void(uint256 streamId)
         external
         payable
--- a/src/SablierFlow.sol
+++ b/src/SablierFlow.sol
@@ -406,8 +406,8 @@ contract SablierFlow is
 
     /// @inheritdoc ISablierFlow
     function transferFrom(IERC20 token, address to, uint128 amount) external payable {
-        // Interaction: transfer the amount using ERC20's `transferFrom` function.
-        token.transferFrom(msg.sender, to, amount);
+        // Interaction: transfer the amount.
+        token.transferFrom({ from: msg.sender, to: to, value: amount });
     }
 
     /// @inheritdoc ISablierFlow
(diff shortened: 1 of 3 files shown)


src/SablierFlow.sol
```diff
@@ -404,6 +404,12 @@ contract SablierFlow is
         _deposit(streamId, amount);
     }
 
+    /// @inheritdoc ISablierFlow
+    function transferFrom(IERC20 token, address to, uint128 amount) external payable {
+        // Interaction: transfer the amount using ERC20's `transferFrom` function.
+        token.transferFrom(msg.sender, to, amount);
```
andreivladbrg, line 410: should we use `token.safeTransferFrom`?

if yes, would the function name would be misleading?

smol-ninja: Advantage of using `safeTransferFrom` here?

Re. the function name: see https://github.com/sablier-labs/flow/issues/361#issuecomment-2652130968

src/interfaces/ISablierFlow.sol
```diff
@@ -356,11 +356,11 @@ interface ISablierFlow is
     /// @param amount The deposit amount, denoted in token's decimals.
     function restartAndDeposit(uint256 streamId, UD21x18 ratePerSecond, uint128 amount) external payable;
 
-    /// @notice Transfers tokens from `msg.sender` to the `to` address using `transferFrom` function of ERC20 contract.
+    /// @notice Transfers ERC-20 tokens from the caller to the `to` address using `token`.
```
smol-ninja, line 359: using `token` is confusing.

IMO "Transfers ERC-20 tokens from the caller to the `to` address" is enough or "Transfers ERC-20 tokens from the caller to the `to` address using `transferFrom` function of ERC20 contract" as before.

src/SablierFlow.sol
andreivladbrg: > Advantage of using `safeTransferFrom` here?

security reasons i guess

smol-ninja: `safeTransferFrom` is useful if we want to check the return value. `transferFrom` always return `true` regardless of token transfer or not. In this case, since we are not checking the return value, there is no need to use `safeTransferFrom`.

src/interfaces/ISablierFlow.sol
andreivladbrg: > using token is confusing

`token` is a param and we use "``", i believe it is not

src/SablierFlow.sol
andreivladbrg: fair point, let's keep `transferFrom`

namewise, i believe it's fine to keep it like this

andreivladbrg (approved):
r_downey_thumbs_up

PR #392, opened 2025-02-12
feat: create with start time

### LINKED ISSUE #351: Create stream with the snapshot time in the past or the future
Add a new function `createWithTimestamp` or `createWithStartTime` (name TBD) that enables users to create a stream with a snapshot time either in the past or the future.

Note that as discussed in https://github.com/sablier-labs/flow/discussions/364, this feature requires getting rid of the $\text{snapshot time} \le \text{block time}$ invariant.

As indicated by Derek below, this feature would be a super useful feature:

andreivladbrg:
Closes #351 

### Changes

- add `startTime` param in `create` function
- add `PENDING` status
- include the `snapshotTime` in the Create event
- allow adjust only with `rps` values greater than zero
- don't allow to create a `PENDING` stream with `rps = 0`
- update the snapshot variables in `_adjust` only if the `st` is in the past
- update create function tests accordingly to the new param
    - add total debt assert after creation
- rename constant to `ONE_MONTH_SINCE_CREATE`
- use 0 as the default start time
- remove `invariant_BlockTimestampGeSnapshotTime` and add `invariant_PendingStream_RpsGt0_TdEq0`

**Note:** 
1. I have not updated the `void` function snapshot time update similar to the `_adjust` function for two reasons: (i) the stream becomes frozen, so it does not matter whether the snapshot time went backwards in time and (ii) I think it’s a good idea to save the timestamp when the stream got voided
2. I have not added the start time as a fuzz parameter in the fork test to avoid overcomplicating things.


src/interfaces/ISablierFlow.sol
```diff
@@ -190,6 +195,7 @@ interface ISablierFlow is
     /// @param recipient The address receiving the tokens.
     /// @param ratePerSecond The amount by which the debt is increasing every second, denoted as a fixed-point number
     /// where 1e18 is 1 token per second.
+    /// @param startTime The timestamp when the stream begins accumulating debt.
```
smol-ninja, line 198: Why not just say "The timestamp when the stream starts" to make it simple? The debt reference is already covered in "Notes". There is no where in `create`'s natspec where we are referring to the stream start time.

So my suggestion is that in `@notice` and `@params`, can we use "stream start" reference and then in "Notes" (which is under `@dev`), we can add the snapshot time reference (which you already did for 0 value).

I have also changed my mind about renaming `snapshot time` to `checkpoint`. I think we can use it (though as per [ChatGPT its unusual](https://chatgpt.com/share/67b076a4-44b0-8000-bcac-2dbb6605c20d)) but works too.

src/SablierFlow.sol
```diff
@@ -490,10 +492,15 @@ contract SablierFlow is
         uint256 blockTimestamp = block.timestamp;
         uint256 snapshotTime = _streams[streamId].snapshotTime;
 
+        // If the snapshot time is in the future, return zero.
+        if (snapshotTime > blockTimestamp) {
```
smol-ninja, line 496: Should be `>=` because it should return 0 when the equality holds.

src/SablierFlow.sol
```diff
@@ -594,6 +606,16 @@ contract SablierFlow is
             revert Errors.SablierFlow_InvalidTokenDecimals(address(token));
         }
 
+        // If the start time is zero, set the snapshot time to `block.timestamp`.
+        uint40 snapshotTime;
+        if (startTime == 0) {
+            snapshotTime = uint40(block.timestamp);
+        }
+        // Otherwise, the snapshot time is the start time.
```
smol-ninja, line 614:
```suggestion
        // Otherwise, set the snapshot time to the start time.
```

src/SablierFlow.sol
```diff
@@ -490,10 +492,15 @@ contract SablierFlow is
...
+        if (snapshotTime > blockTimestamp) {
+            return 0;
+        }
+
         uint256 ratePerSecond = _streams[streamId].ratePerSecond.unwrap();
 
-        // Check:if the rate per second is zero or the `block.timestamp` is less than the `snapshotTime`.
-        if (ratePerSecond == 0 || blockTimestamp <= snapshotTime) {
+        // Check: if the rate per second is zero.
+        if (ratePerSecond == 0) {
```
smol-ninja, line 503: What do you think about merging this condition with the above?

```solidity
if (snapshotTime > blockTimestamp || ratePerSecond == 0)
```

I think its cleaner to merge conditions together that return the same value (`0`).

src/interfaces/ISablierFlow.sol
```diff
@@ -174,11 +176,14 @@ interface ISablierFlow is
     /// per second.
     function adjustRatePerSecond(uint256 streamId, UD21x18 newRatePerSecond) external payable;
```
smol-ninja, line 177: `Notes` section of `adjustRatePerSecond` should be updated to consider the case when snapshot time is in the future.

smol-ninja (changes requested):
I have only reviewed the source code for now because of the following feedback:

1. When snapshot time is in the future, during the period until that point in time, the stream is neither streaming nor paused (pause implies rps = 0). So we need a new `Status` called `PENDING`.

2. Several natspec update such as replacing `Returns the amount of debt accrued since the snapshot time until now` with `Returns the amount of debt accrued at the snapshot time` such as in `ongoingDebtScaledOf`. We can consider snapshot time as an absolute timestamp and remove references such as "since the snapshot time until now" since it can be anywhere now. 

3. Alternatively, we can also add a note stating that if `snapshot time` is in the future, it will return 0.

4. From the design point of view, `pause` should set snapshot time to current timestamp, similar to `void`. Pausing is a sender action and should update st.

5. Should we provide a new function to start a scheduled Flow stream earlier than scheduled time? Flow is supposed to be flexible and it would be quite useful (cc @razgraf @PaulRBerg). May be the employees joins earlier than expected.

6. `restart` should revert if stream has not been started.

7. I agree with you to allow void on a stream not started and updating st to current time.

src/SablierFlow.sol
andreivladbrg: IMO it is clearer with individual if blocks

it is easier to follow - improves readability

smol-ninja: > improves readability

The combined one improves readability as well. Its pretty common to use bool gates in programming. Not sure why you dont like it, but its not something that nobody uses.

andreivladbrg:
> When snapshot time is in the future, during the period until that point in time, the stream is neither streaming nor paused (pause implies rps = 0). So we need a new Status called PENDING.

correct, we have spoken about this on slack, but i forgot to comment this here.
i also missed to ask about how we should name the `st`.

> Several natspec update such as replacing Returns the amount of debt accrued since the snapshot time until now with Returns the amount of debt accrued at the snapshot time such as in ongoingDebtScaledOf. We can consider snapshot time as an absolute timestamp and remove references such as "since the snapshot time until now" since it can be anywhere now.

what do you say if we add: "unless the status is `PENDING`" - then the statements hold true.

> Alternatively, we can also add a note stating that if snapshot time is in the future, it will return 0.
        
i don't like this idea. we better rename the var

> From the design point of view, pause should set snapshot time to current timestamp, similar to void. Pausing is a sender action and should update st.

i disagree, pause is more like a "temporarily" pause so i believe void is different. also, even if we set it to current timestamp in pause, what are we going to do in adjust when you pass a `rps = 0`

> restart should revert if stream has not been started.

why? disagree.

---

what if we go back and name this variable how was initially called: `lastTimeUpdate` or `lastUpdatedTime` 
IMO it makes total sense and it is a clear name.

smol-ninja:
> i also missed to ask about how we should name the st.

As I mentioned here, I am fine calling it snapshot time. Snapshots can be in future so it sounds alright.

> what do you say if we add: "unless the status is PENDING" - then the statements hold true

Status is not the storage state but something we compute in runtime. While snapshot time is a state variable so referring to it would be more useful for developers.

> i don't like this idea. we better rename the var

Liking is a personal preference :). To me, renaming the variable is fine too as long as it makes it clear. I suggested because finding a good name for variable is hard. We use snapshot as an important keyword so renaming snapshot time may break other commends too. Feel free to take a call on this.

> why? disagree

What does the restart do when status is pending? Do nothing? Similar to why we revert on pause when a stream is paused, when we can also do nothing, restarting a scheduled stream does not make sense because its never paused.

> what if we go back and name this variable how was initially called: lastTimeUpdate or lastUpdatedTime

Personally, I never liked that name. Snapshot time makes the Flow docs very explainable. Its the time when debt snapshots are taken.

src/SablierFlow.sol
andreivladbrg: as gpt said: https://chatgpt.com/share/67b0a126-4ab8-8013-a941-5bb3d15c7fb8

"if you want clarity **Version 2 is better**" - that's what i said above (clearer)

smol-ninja: Even though I still disagree (and also disagree with ChatGPT here), the clarity difference is as tiny as the gas difference between using `>` and `>=`, in my opinion. A smart developers (the audience here) can equally see through them.

But lets go with version 2.

andreivladbrg:
@smol-ninja i have pushed some new commits following the decision we have during the call

also, the invariant i came up with, was this: `invariant_PendingStream_RpsGt0_TdEq0` which i believe covers all possible scenarios

src/SablierFlow.sol
```diff
@@ -126,6 +126,11 @@ contract SablierFlow is
 
     /// @inheritdoc ISablierFlow
     function statusOf(uint256 streamId) external view override notNull(streamId) returns (Flow.Status status) {
+        // Check: the stream has started.
```
smol-ninja, line 129: Its not a check. And I don't think `check` makes sense in view functions. We are not reverting anything here.

src/SablierFlow.sol
```diff
@@ -198,6 +203,11 @@ contract SablierFlow is
         onlySender(streamId)
         updateMetadata(streamId)
     {
+        // Check: the rate per second is not zero.
```
smol-ninja, line 206:
```suggestion
        // Check: the new rate per second is not zero.
```

src/SablierFlow.sol
```diff
@@ -198,6 +203,11 @@ contract SablierFlow is
...
+        if (newRatePerSecond.unwrap() == 0) {
+            revert Errors.SablierFlow_RatePerSecondZero();
```
smol-ninja, line 208: We should rename it to `Errors.SablierFlow_NewRatePerSecondZero()` to avoid confusion with the current rate per second.

src/SablierFlow.sol
```diff
@@ -556,16 +573,21 @@ contract SablierFlow is
             revert Errors.SablierFlow_RatePerSecondNotDifferent(streamId, newRatePerSecond);
```
smol-ninja, line 573: Should we move this check into `adjustRatePerSecond`? I dont like the fact that we have one check in `adjustRatePerSecond` and another in `_adjustRatePerSecond`.

src/SablierFlow.sol
```diff
@@ -594,6 +617,21 @@ contract SablierFlow is
             revert Errors.SablierFlow_InvalidTokenDecimals(address(token));
         }
 
+        // If the start time is zero, set the snapshot time to `block.timestamp`.
+        uint40 snapshotTime;
+        if (startTime == 0) {
+            snapshotTime = uint40(block.timestamp);
```
smol-ninja, line 623: Since `block.timestamp` is used twice, lets load it into a variable.

src/SablierFlow.sol
```diff
@@ -594,6 +617,21 @@ contract SablierFlow is
...
+        }
+        // Otherwise, set the snapshot time to the start time.
+        else {
+            snapshotTime = startTime;
+        }
+
+        // Check: the rate per second is not zero if the start time is in the future.
+        if (startTime > block.timestamp && ratePerSecond.unwrap() == 0) {
```
smol-ninja, line 631: Why not move this over if else to have all the checks together?

src/libraries/Errors.sol
```diff
@@ -52,6 +55,9 @@ library Errors {
     /// @notice Thrown when trying to restart a stream that is not paused.
     error SablierFlow_StreamNotPaused(uint256 streamId);
 
+    /// @notice Thrown when trying to pause a stream that has not started.
+    error SablierFlow_StreamNotStarted(uint256 streamId, uint40 snapshotTime);
```
smol-ninja, line 59: Should we call it `startTime` since the error name is `StreamNotStarted`? Calling it `startTime` makes it more intuitive what the variable represents.

src/SablierFlow.sol
```diff
@@ -556,16 +573,21 @@ contract SablierFlow is
...
-        uint256 ongoingDebtScaled = _ongoingDebtScaledOf(streamId);
+        uint40 blockTimestamp = uint40(block.timestamp);
 
-        // Update the snapshot debt only if the stream has ongoing debt.
-        if (ongoingDebtScaled > 0) {
-            // Effect: update the snapshot debt.
-            _streams[streamId].snapshotDebtScaled += ongoingDebtScaled;
-        }
+        // Update the snapshot variables only if the snapshot time is not in the future.
+        if (_streams[streamId].snapshotTime < blockTimestamp) {
```
smol-ninja, line 579: `<=` since its OK to upadte values when the two values match.

src/SablierFlow.sol
```diff
@@ -710,12 +755,12 @@ contract SablierFlow is
...
     /// @dev See the documentation for the user-facing functions that call this internal function.
     function _restart(uint256 streamId, UD21x18 ratePerSecond) internal {
-        // Check: the stream is not paused.
+        // Check: the stream is paused.
         if (_streams[streamId].ratePerSecond.unwrap() != 0) {
             revert Errors.SablierFlow_StreamNotPaused(streamId);
         }
 
-        // Checks and Effects: update the rate per second and the snapshot time.
+        // Checks and Effects: restart the stream by adjusting the rate per second to a value greater than zero.
```
smol-ninja, line 763:
```suggestion
        // Checks and Effects: restart the stream by adjusting the rate per second to the value provided.
```

You have already checked for zero value so its not zero anymore.

TECHNICAL-DOC.md
```diff
@@ -55,40 +55,40 @@ can only withdraw the available balance.
 
 ## Invariants
 
-1. for any stream, $st \le now$
```
smol-ninja, line 58: We should keep this invariant with a mild modification:

For any non-pending stream, $st \le now$

TECHNICAL-DOC.md
```diff
@@ -55,40 +55,40 @@ can only withdraw the available balance.
...
+1. for a given token:
 
    - $\sum$ stream balances = aggregate balance
    - token.balanceOf(SablierFlow) $`\ge \sum`$ stream balances
    - $\sum$ stream balances = $\sum$ deposited amount - $\sum$ refunded amount - $\sum$ withdrawn amount
 
-3. For a given token, token.balanceOf(SablierFlow) $\ge$ flow.aggregateBalance(token)
+2. For a given token, token.balanceOf(SablierFlow) $\ge$ flow.aggregateBalance(token)
 
-4. snapshot time should never decrease
```
smol-ninja, line 68: Even for voided stream, why would the snapshot time decrease? So I think this invariant is true for all non-pending streams.

for any non-pending stream the snapshot time should never decrease.

TECHNICAL-DOC.md
```diff
@@ -55,40 +55,40 @@ can only withdraw the available balance.
...
 
-11. if $` ud = 0 \implies cd = td`$
+10. if $` ud = 0 \implies cd = td`$
 
-12. $bal = ra + cd$
+11. $bal = ra + cd$
 
-13. for any non-voided stream, if $rps \gt 0 \implies isPaused = false$ and Flow.Status is either STREAMING_SOLVENT or
-    STREAMING_INSOLVENT.
+12. for any non-voided stream, if $rps \gt 0 \implies isPaused = false$ and Flow.Status is either PENDING,
```
smol-ninja, line 84: Isn't it true for all streams, even if they are voided? Because a stream cannot be voided if rps > 0, this applies to all imo.

We should remove "for any non-voided stream"

DIAGRAMS.md
```diff
@@ -5,10 +5,12 @@
 
 ## Statuses
 
-There are two types of streams: `STREAMING`, when debt is actively accruing, and `PAUSED`, when debt is not accruing:
+There are three types of streams: `PENDING` when stream is scheduled for a future start, `STREAMING`, when debt is
+actively accruing, and `PAUSED`, when debt is not accruing:
 
 | Type        | Status                | Description                                                                             |
 | ----------- | --------------------- | --------------------------------------------------------------------------------------- |
+| `PENDING`   | `PENDING`             | A stream that has not started accumulating debt.                                        |
```
smol-ninja, line 13: A stream that has not been started. Because a stream that has not started accumulating debt does not necessary have pending status if st <= now and rps = 0.

DIAGRAMS.md
```diff
@@ -44,9 +46,13 @@ stateDiagram-v2
     Streaming --> VOIDED : void
 
     NULL --> Streaming : create (rps > 0)
+    NULL --> Pending : create (rps > 0)
```
smol-ninja, line 49: should we be more explicit with `create (rps > 0, st > now)`? By looking at the status, since `create (rps > 0)` can lead to two statuses, it does not add much clarity as what made them different.

DIAGRAMS.md
```diff
@@ -103,6 +110,7 @@ flowchart LR
 
     CR -- "update rps<br/>update st" --> NULL
 
+    ADJRPS -- "update rps" --> PND
```
smol-ninja, line 113: This does not look correct. the LHS should be CREATE

DIAGRAMS.md
```diff
@@ -116,12 +124,13 @@ flowchart LR
 
     RST -- "update rps<br/>update st" --> PSED
 
+    VD -- "update rps (0)<br/>update st" --> PND
```
smol-ninja, line 127: How can void lead to pending?

tests/utils/Constants.sol
```diff
@@ -30,9 +30,10 @@ abstract contract Constants is CommonConstants {
     // Time
     uint40 internal constant FEB_1_2025 = 1_738_368_000;
     uint40 internal constant ONE_MONTH = 30 days; // "30/360" convention
-    uint40 internal constant ONE_MONTH_SINCE_START = FEB_1_2025 + ONE_MONTH;
+    uint40 internal constant ONE_MONTH_SINCE_CREATE = FEB_1_2025 + ONE_MONTH;
     // Solvency period is 49999999.999999 seconds.
     uint40 internal constant SOLVENCY_PERIOD = uint40(DEPOSIT_AMOUNT_18D / RATE_PER_SECOND_U128); // ~578 days
+    uint40 internal constant START_TIME_ZERO = 0;
```
smol-ninja, line 36: Lets call it `ZERO` so that we can use whenever we need 0.

tests/integration/fuzz/pause.t.sol
```diff
@@ -2,13 +2,40 @@
...
 import { IERC4906 } from "@openzeppelin/contracts/interfaces/IERC4906.sol";
+import { UD21x18, ud21x18 } from "@prb/math/src/UD21x18.sol";
 
 import { ISablierFlow } from "src/interfaces/ISablierFlow.sol";
 import { Errors } from "src/libraries/Errors.sol";
 
 import { Shared_Integration_Fuzz_Test } from "./Fuzz.t.sol";
 
 contract Pause_Integration_Fuzz_Test is Shared_Integration_Fuzz_Test {
+    function testFuzz_RevertGiven_NotStarted(
```
smol-ninja, line 13: Even though the `/// @dev` comment missing, I will let it go because adding dev does not offer much to it.

tests/integration/Integration.t.sol
```diff
@@ -29,6 +29,7 @@ abstract contract Integration_Test is Base_Test {
             sender: users.sender,
             recipient: users.recipient,
             ratePerSecond: ratePerSecond,
+            startTime: START_TIME_ZERO,
```
smol-ninja, line 32: Since 0 is a special case, my suggestion would be to have `block.timestamp` as start time for default tests.

tests/fork/Flow.t.sol
smol-ninja, line 1: Why not add `startTime` in the params?

tests/integration/fuzz/pause.t.sol
smol-ninja, line 1: In fuzz tests, we should update `useFuzzedStreamOrCreate` to create some pending streams so that other operations can be tested when stream status is pending. We do not have those tests yet.

tests/invariant/handlers/FlowHandler.sol
```diff
@@ -165,6 +164,9 @@ contract FlowHandler is BaseHandler {
         // Paused streams cannot be paused again.
         vm.assume(!flow.isPaused(currentStreamId));
 
+        // The stream must not be PENDING.
+        vm.assume(flow.getSnapshotTime(currentStreamId) < getBlockTimestamp());
```
smol-ninja, line 168: Should be `<=`

tests/invariant/handlers/FlowCreateHandler.sol
```diff
@@ -49,9 +49,19 @@ contract FlowCreateHandler is BaseHandler {
         address sender;
         address recipient;
         uint128 ratePerSecond;
+        uint40 startTime;
         bool transferable;
     }
 
+    /// @dev Struct to prevent stack too deep error.
+    struct Vars {
```
smol-ninja, line 57: Do we really need the full struct? Can defining `streamId` in the storage solve this?

smol-ninja (changes requested):
Fantastic work. It clearly wasn't as easy as we thought first. Please find my feedback below.

tests/fork/Flow.t.sol
andreivladbrg: mentioned in the OP: "I have not added the start time as a fuzz parameter in the fork test to avoid overcomplicating things"

happy to add it in the fuzz tests

tests/integration/Integration.t.sol
andreivladbrg: in the real world, 0 will be used instead of `block.timestamp` in the UI.

the reason for not using it is that if the UI passes `block.timestamp` and the tx is executed in X num of seconds, it would create a stream with a start time in the past

so i'd suggest having the default value what will be more often used

src/interfaces/ISablierFlow.sol
```diff
@@ -330,7 +340,7 @@ interface ISablierFlow is
     ///
     /// Requirements:
     /// - Must not be delegate called.
-    /// - `streamId` must not reference a null, or a voided stream.
+    /// - `streamId` must not reference a null, a not paused or a voided stream.
```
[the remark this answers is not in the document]
andreivladbrg: i know and you know it.
but i believe now we should be super explicit in natspec

src/interfaces/ISablierFlow.sol
```diff
@@ -277,7 +287,7 @@ interface ISablierFlow is
     ///
     /// Requirements:
     /// - Must not be delegate called.
-    /// - `streamId` must not reference a null or an already paused stream.
+    /// - `streamId` must not reference a null, a not started or an already paused stream.
```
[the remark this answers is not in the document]
andreivladbrg: i wanted to say initially "pending" but i remembered your feedback 😅:

image

happy to change it

src/interfaces/ISablierFlow.sol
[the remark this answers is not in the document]
smol-ninja: OK then Id suggest to reframe it like following:

`streamId` must not reference a null, must not be paused or voided.

instead of using "a not paused"

src/SablierFlow.sol
andreivladbrg: agree

src/SablierFlow.sol
andreivladbrg: hmm, the reason for not saying "New" is because we can use this error in both create and adjust, so i would leave like this. 
usually checks are made on input params, so i believe is fine leaving it like this

src/SablierFlow.sol
andreivladbrg: > Should we move this check into adjustRatePerSecond

we need this check in restart as well (in case you want to pass a rate per second of zero, this will revert with `SablierFlow_RatePerSecondNotDifferent`

> I dont like the fact that we have one check in adjustRatePerSecond and another in _adjustRatePerSecond 

as you said above "Liking is a personal preference :)."  😛

the fact we don't need more errors and we can solely have a single point of "failure" (internal `_adjust`) is a better design

src/SablierFlow.sol
andreivladbrg: the reason i didn't add it in memory was because this line is inside the if block.
so if `startTime` is zero, then, yes it is more gas efficient. if it is not, then is less gas efficient 

but, given we are talking about a tiny bit, i ok with both of these options

src/SablierFlow.sol
andreivladbrg: i don't understand, can you give the exact code you are talking about?

src/SablierFlow.sol
smol-ninja: Fair point in reusability. But we have a pattern where we focus more on bespoke error names over reusability. That was my rationale behind the feedback. 

If we want to shift the focus on reusability then we can merge several other errors too such as `SablierFlow_StreamBalanceZero(uint256 streamId)`, `SablierFlow_RefundAmountZero(uint256 streamId);`, `SablierFlow_DepositAmountZero(uint256 streamId);` into one called `SablierFlow_AmountZero(uint256 streamId);`

TECHNICAL-DOC.md
andreivladbrg: agree

src/SablierFlow.sol
smol-ninja:
```solidity
        // Check: the token decimals are not greater than 18.
        if (tokenDecimals > 18) {
            revert Errors.SablierFlow_InvalidTokenDecimals(address(token));
        }

        // Check: the rate per second is not zero if the start time is in the future.
        if (startTime > block.timestamp && ratePerSecond.unwrap() == 0) {
            revert Errors.SablierFlow_RatePerSecondZero();
        }

        // If the start time is zero, set the snapshot time to `block.timestamp`.
        uint40 snapshotTime;
        if (startTime == 0) {
            snapshotTime = uint40(block.timestamp);
        }
        // Otherwise, set the snapshot time to the start time.
        else {
            snapshotTime = startTime;
        }

        // Load the stream ID.
        streamId = nextStreamId;

        // Effect: create the stream.
        _streams[streamId] = Flow.Stream({
            balance: 0,
            isStream: true,
            isTransferable: transferable,
            isVoided: false,
            ratePerSecond: ratePerSecond,
            sender: sender,
            snapshotDebtScaled: 0,
            snapshotTime: snapshotTime,
            token: token,
            tokenDecimals: tokenDecimals
        });
```

TECHNICAL-DOC.md
andreivladbrg: > Because a stream cannot be voided if rps > 0

that's incorrect. you can void any kind of stream

the only things that must be respected are: (i) null stream (ii) already voided

src/libraries/Errors.sol
smol-ninja: Fair point.

DIAGRAMS.md
andreivladbrg: this diagrams is for what functions can be called based on their status :))

https://github.com/sablier-labs/flow/blob/21f29f8acd125742deef33ba84db66b6a30e7dd6/DIAGRAMS.md#function-calls

and void can be called on pending status

image

tests/integration/fuzz/pause.t.sol
andreivladbrg: thank you for "letting me go" 😅

src/SablierFlow.sol
smol-ninja: Agree. In that case, lets change the comment to `Update the snapshot variables only if the snapshot time is in the past` because "not in future" implies present too.

tests/invariant/handlers/FlowCreateHandler.sol
andreivladbrg: i added for all variables because `createAndDeposit` was causing the error.
but if only declaring the `streamId` as internal would fix stack too deep error, i am ok to remove the struct.

src/SablierFlow.sol
smol-ninja: Oh right. Even in that case "restart the stream by adjusting the rate per second to a value greater than zero" is ambiguous. The new value is provided as the input so what does it mean to say "a value greater than zero". It either reverts if its zero or restarts the stream if its not zero.

TECHNICAL-DOC.md
smol-ninja: > that's incorrect. you can void any kind of stream

I meant to say "a stream cannot have voided status if rps > 0". So with the corrected statement, for any stream, if rps > 0, implies ispaused = false and status is either pending, or streaming.

src/types/DataTypes.sol
```diff
@@ -7,16 +7,20 @@ import { UD21x18 } from "@prb/math/src/UD21x18.sol";
...
+    /// 1. Pending: when stream is scheduled for a future start.
+    /// 2. Streaming: when the total debt is increasing.
+    /// 3. Paused: when the total debt is not increasing.
     ///
-    /// @custom:value0 STREAMING_SOLVENT Streaming stream when there is no uncovered debt.
-    /// @custom:value1 STREAMING_INSOLVENT Streaming stream when there is uncovered debt.
-    /// @custom:value2 PAUSED_SOLVENT Paused stream when there is no uncovered debt.
-    /// @custom:value3 PAUSED_INSOLVENT Paused stream when there is uncovered debt.
-    /// @custom:value4 VOIDED Paused stream with no uncovered debt and it cannot be restarted.
+    /// @custom:value0 PENDING A stream that has not started accumulating debt.
```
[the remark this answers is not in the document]
andreivladbrg: > /// @custom:value0 PENDING A stream with snapshot time in the future.

all the other `@custom` natspec comments for the different statuses mention soemthing about "debt" so we should also mention this for PENDING

what about this: "A stream with an rps greater than zero that has not yet started accumulating debt"

src/interfaces/ISablierFlow.sol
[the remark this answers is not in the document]
andreivladbrg: > But now I think calling it "snapshot time in future" is much less explicit than calling it "pending". Not pending as status but as the definition that snapshot time is in the future.

thank you, agree that "pending" is better

tests/fork/Flow.t.sol
smol-ninja: OK if its in fuzz, not having it in fork is fine.

tests/invariant/handlers/FlowCreateHandler.sol
smol-ninja: Lets give that a shot then. If it does not work, then we have no choice but to introduce this struct.

src/types/DataTypes.sol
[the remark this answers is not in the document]
smol-ninja: I assume you want to follow consistency here. But `A stream with an rps greater than zero that has not yet started accumulating debt` vs `A stream with snapshot time in the future`, the latter is more crystal clear. So what if we have to use snapshot time instead of debt here? Debt in other values make sense, but in the context of pending, only snapshot time is sufficient.

src/SablierFlow.sol
andreivladbrg: i said that 0 will be used instead of passing `block.timestamp` as a param, and not that this will be the majority of cases. maybe `PENDING` or a start time in the past will be used more often—it’s impossible to know in advance

but as said above. i agree with the change

src/SablierFlow.sol
andreivladbrg: oh, right, yes

agree

src/interfaces/ISablierFlow.sol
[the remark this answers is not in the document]
smol-ninja: To me, it sounds better than `streamId` must not reference a null, a not paused or a voided stream.`. Having explicit "must not" has better readability.

[Confirmed with ChatGPT](https://chatgpt.com/share/67b5dc39-b9a4-8000-b56d-65a214fc20b9).

src/SablierFlow.sol
andreivladbrg: > lets change the comment to Update the snapshot variables only if the snapshot time is in the past because "not in future" implies present too

good catch

DIAGRAMS.md
andreivladbrg: responded below

src/interfaces/ISablierFlow.sol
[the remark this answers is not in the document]
andreivladbrg: > [Confirmed with ChatGPT](https://chatgpt.com/share/67b5dc39-b9a4-8000-b56d-65a214fc20b9).

[refuted with ChatGPT](https://chatgpt.com/share/67b5dfbe-8690-8013-8876-611c83b1ccb2)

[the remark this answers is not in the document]
smol-ninja: You refuted with ChatGPT 4o I think which is an outdated model. I used the same prompt as yours and model o3-mini think second is clearer.

https://chatgpt.com/share/67b5e206-69fc-8000-9708-3bb857b38d8e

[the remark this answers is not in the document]
smol-ninja: Same with `mini-high` which is the most advanced I think 

https://chatgpt.com/share/67b5e206-69fc-8000-9708-3bb857b38d8e

TECHNICAL-DOC.md
andreivladbrg: > Even for voided stream, why would the snapshot time decrease? 

you can switch from `pending --> voided` and then inv won't hold, as `st` went backwards 

a pending stream means `st = constant`

DIAGRAMS.md
andreivladbrg: IIRC i tried this and it didn't look as nice. if you have a strong preference let's do it separately

andreivladbrg:
@smol-ninja  addressed most of the feedback(expect for fuzz and inv tests) in e19fbc06269c35808c208e052d747122e80b87f9 and 343fb3da8cfb6f52c8030e27b52d22a77195d3a8

lmk if it looks good

TECHNICAL-DOC.md
smol-ninja: That is why I said "for any non-pending stream". When you switch from pending to voided, the invariant does not apply on the initial state (pending).

DIAGRAMS.md
smol-ninja: OK.

TECHNICAL-DOC.md
andreivladbrg: right, so we can remove "for any non-voided stream"

tests/integration/fuzz/pause.t.sol
andreivladbrg: i thought more about this, and having pending streams created wouldn’t expand the tests into more "hot" scenarios, as the calculation functions (td, ud etc.) will always return 0. so, IMO we should fuzz the start time but only for values less than `block.timestamp`. wdyt?

smol-ninja: I wasn't clear in my comment. What I had in my mind is to create a few extra streams through `_setupStreamsWithAllDecimals` as PENDING. Currently it creates 20 streams, we can extend it to 22 streams with 2 as PENDING.

Then whenever `useFuzzedStreamOrCreate` is called, if the stream ID does not exist, it always create a default stream (which is not pending). But if the stream ID exists, there are 2/22 chances that a pending stream will be a part of the test. So atleast, very rare, but we will be covering pending streams through fuzz tests.

Wdyt?

andreivladbrg: ok, i get it now. 

>  we can extend it to 22 streams with 2 as PENDING.

this sounds good

tests/invariant/handlers/FlowCreateHandler.sol
```diff
@@ -49,9 +49,12 @@ contract FlowCreateHandler is BaseHandler {
         address sender;
         address recipient;
         uint128 ratePerSecond;
+        uint40 startTime;
         bool transferable;
     }
 
+    uint256 internal streamId;
```
smol-ninja, line 56: Why not move it under `VARIABLES` section at the top?

smol-ninja (changes requested):
Generally looks good. Just three comments below plus concrete tests for pending streams as we discussed.

TECHNICAL-DOC.md
```diff
@@ -83,17 +83,19 @@ can only withdraw the available balance.
...
 
-13. for any non-voided stream, if $rps \gt 0 \implies isPaused = false$ and Flow.Status is either STREAMING_SOLVENT or
+13. if $rps \gt 0 \implies isPaused = false$ and Flow.Status is either PENDING, STREAMING_SOLVENT or
     STREAMING_INSOLVENT.
 
 14. for any non-voided stream, if $rps = 0 \implies isPaused = true$ and Flow.Status is either PAUSED_SOLVENT or
     PAUSED_INSOLVENT.
 
-15. if $isPaused = true \implies rps = 0$
+15. for any PENDING stream, $\implies rps > 0$ and $td = 0$
```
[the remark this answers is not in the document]
andreivladbrg: i'd keep the current version, consistent with all invariants

tests/invariant/Flow.t.sol
```diff
@@ -227,6 +232,22 @@ contract Flow_Invariant_Test is Base_Test, StdInvariant {
         }
     }
 
+    /// @dev For any stream, if the snapshot time is greater than the current time, then the rate per second should be
```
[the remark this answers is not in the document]
andreivladbrg: the comment says: `if the snapshot time is greater than the current time` - which is the definition of "PENDING" stream

but i will keep the terminology from technical-doc

TECHNICAL-DOC.md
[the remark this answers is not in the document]
smol-ninja: How is it consistent? `implies` is used with an `if`. if a is true implies this and that.

So the correct statement would be If a stream is PENDING, $\implies rps > 0$ and $td = 0$. or for a PENDING stream, $rps > 0$ and $td = 0$

tests/invariant/Flow.t.sol
```diff
@@ -129,16 +116,34 @@ contract Flow_Invariant_Test is Base_Test, StdInvariant {
...
+                assertGe(
+                    getBlockTimestamp(),
+                    flow.getSnapshotTime(streamId),
+                    "Invariant violation: pending stream with block timestamp < snapshot time"
+                );
+            }
+        }
+    }
+
     /// @dev For any stream, the snapshot time should be greater than or equal to the previous snapshot time.
```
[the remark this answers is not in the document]
andreivladbrg: > This should match the invariant statement from technical doc which is "for any non-voided stream the snapshot time should never decrease

yep

> Also, can we move it above invariant_NonPending_BlockTimestampGeSnapshotTime so that its in the same order as written in the technical doc?


the inv from this file should be ordered alphabetically

TECHNICAL-DOC.md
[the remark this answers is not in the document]
andreivladbrg: oh, mb, we should: `for any PENDING stream, $rps > 0$ and $td = 0$`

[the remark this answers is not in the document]
smol-ninja: Yes. That would be better.

tests/invariant/Flow.t.sol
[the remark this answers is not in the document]
andreivladbrg: nvm, let's keep the order from technical doc, its hard to know in advance the alphabetic order

andreivladbrg:
as discussed on slack, i’ve added concrete tests for `adjust` and `void` in the case of `PENDING` streams. also, i’ve added tests for the ongoing debt function, as all the other functions use it underneath—so if it’s zero for it, that should be enough for the others as well.

and i addressed the feedback left in 92ea7a8c9a1761a812b77374bed544f66ee7160c.
lmk if there’s anything else.

smol-ninja (approved):

PR #397, opened 2025-02-25
refactor: bump evm-utils

smol-ninja:


andreivladbrg (approved):
👍🏻

tests/utils/BaseScript.t.sol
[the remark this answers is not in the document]
andreivladbrg: 540cb7f27e258ed93d1fdd661d060b6641782c9c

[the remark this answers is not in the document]
smol-ninja: Yeah agree.

script/DeployDeterministicFlow.s.sol
```diff
@@ -11,8 +11,9 @@ import { SablierFlow } from "src/SablierFlow.sol";
 /// @dev Reverts if the contract has already been deployed.
 contract DeployDeterministicFlow is BaseScript {
     function run() public broadcast returns (SablierFlow flow, FlowNFTDescriptor nftDescriptor) {
+        address initialAdmin = protocolAdmin();
         bytes32 salt = constructCreate2Salt();
         nftDescriptor = new FlowNFTDescriptor{ salt: salt }();
-        flow = new SablierFlow{ salt: salt }(protocolAdmin(), nftDescriptor);
+        flow = new SablierFlow{ salt: salt }(initialAdmin, nftDescriptor);
```
smol-ninja, line 17: Why did you add a variable here? What was wrong with using `protocolAdmin()` directly?

andreivladbrg: technically no problem.. but as a practice i don't think its a good idea to call a function that may or may not have additional logic after the first deployment (`FlowNFTDescriptor`)
i'd keep the current version

smol-ninja: OK we can keep your suggestion though I don't agree that its a best practice. 1 Line less and still doing the job.

PR #399, opened 2025-03-05
test: run fork tests without forking the chain during development

smol-ninja:
It required some additional changes because for all the tokens, the fork tests were running in a single run. Therefore, in addition to https://github.com/sablier-labs/lockup/issues/1197, this PR also refactors the fork tests to follow the same pattern as Lockup.

### DIFF
--- a/tests/fork/Flow.t.sol
+++ b/tests/fork/Flow.t.sol
@@ -12,7 +12,7 @@ import { Flow } from "src/types/DataTypes.sol";
 
 import { Fork_Test } from "./Fork.t.sol";
 
-contract Flow_Fork_Test is Fork_Test {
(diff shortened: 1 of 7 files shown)


smol-ninja:
I just realised I could have also added 

```solidity
vm.etch(address(tokens[i]), address(usdc).code);
```

inside the for loop. I will let you decide if you want to go with this refactor or including `vm.etch` in the for loop. I am fine with both.

andreivladbrg (approved):
LGTM

andreivladbrg:
> I just realised I could have also added

yeah, it would work, but i believe it's better to have individual files for each token — that way, in case of a revert, we have full context of the token it reverted with, and shorter fork runs since we run them in parallel.

PR #403, opened 2025-03-13
build: use solc 0.8.29

### LINKED ISSUE #402: Add `multibuild` workflow
Just realized that here we don't have a workflow for multibuild like in lockup/airdrops 

https://github.com/sablier-labs/lockup/blob/main/.github/workflows/multibuild.yml

andreivladbrg:
Closes #402


smol-ninja:
Should we include https://github.com/sablier-labs/flow/issues/402 in this PR as well?

PR #405, opened 2025-03-14
Audit fixes

### LINKED ISSUE #404: Audit fixes
### Non-informational fixes:

- Use `safeTransferFrom` in `SablierFlow.transferFrom`, and consider another name to avoid collision with `ERC721.transferFrom`
- For tokens like `POL` and `CELO`, we won't support them at the UI level

### Informational fixes:

- Add "emit {MetadataUpdate}" in all functions
- At `ISablierFlow#L114` say "If the total debt exceeds the stream balance, it returns 0."
- `ISablierFlow.withdraw` "/// - `to` must be the recipient if `msg.sender` is not the stream's recipient or an approved third party."
- `ISablierFlow.createAndDeposit#startTime` Natspec similar to `ISablierFlow.create#startTime`
- `Errors.SablierFlow_InvalidTokenDecimals` to: "Thrown when trying to create a stream with a token with decimals greater than 18."
- `Helpers.scale/descale`: "`decimals` should be less than or equal to 18"
- Add more documentation about "double-entry" tokens, and update Assumptions, that these tokens are not supported.

---

*(For further explanations, go through each finding)*

**Findings:**
- https://cantina.xyz/code/c9980035-706a-4119-91fe-7ebddaf123e4/findings/2
- https://cantina.xyz/code/c9980035-706a-4119-91fe-7ebddaf123e4/findings/5
- https://cantina.xyz/code/c9980035-706a-4119-91fe-7ebddaf123e4/findings/9
- https://cantina.xyz/code/c9980035-706a-4119-91fe-7ebddaf123e4/findings/12

andreivladbrg:
Closes #404

**Note:** I have not changed the [Errors.sol#L67–L68](https://cantina.xyz/code/c9980035-706a-4119-91fe-7ebddaf123e4/flow/src/libraries/Errors.sol#L67-L68) Natspec, as the current version is consistent with how we refer to the caller as `msg.sender` throughout `ISablierFlow`.


src/libraries/Helpers.sol
```diff
@@ -4,7 +4,8 @@ pragma solidity >=0.8.22;
 /// @title Helpers
 /// @notice Library with helper functions in {SablierFlow} contract.
 library Helpers {
-    /// @dev Descales the provided `amount` from 18 decimals fixed-point number to token's decimals number.
+    /// @notice Descales the provided `amount` from 18 decimals fixed-point number to token's decimals number.
+    /// @dev The `decimals` must not be greater than 18.
```
smol-ninja, line 8: We should also mention what would happen if decimals exceeds 18 which is the result would overflow.

src/libraries/Helpers.sol
```diff
@@ -16,7 +17,9 @@ library Helpers {
         }
     }
 
-    /// @dev Scales the provided `amount` from token's decimals number to 18 decimals fixed-point number.
+    /// @notice Scales the provided `amount` from token's decimals number to 18 decimals fixed-point number.
+    /// @dev The `decimals` must not be greater than 18. The scaled result may overflow `uint256`. If `amount` fits into
```
smol-ninja, line 21: Same here.

Also, the second sentence should be `If amount exceeds `max(uint128)`, the result may overflow `uint256`

this is enough imo

src/SablierFlow.sol
```diff
@@ -417,9 +417,9 @@ contract SablierFlow is
     }
 
     /// @inheritdoc ISablierFlow
-    function transferFrom(IERC20 token, address to, uint128 amount) external payable {
+    function transferTokensFrom(IERC20 token, address to, uint128 amount) external payable {
```
smol-ninja, line 420: Why `transferTokensFrom`? Why not just `transferToken`? Usually, `transferFrom` would imply `from` param

smol-ninja (changes requested):
The following recommendation is missing:

> Consider updating the NatSpec of the `SablierFlow_InvalidTokenDecimals()` error to
`Thrown when trying to create a stream with a token with decimals greater than 18.`

src/SablierFlow.sol
andreivladbrg: Good point, agree.
Wanted to keep the previous terminology, but it makes sense to remove `From`

andreivladbrg:
>> The following recommendation is missing: Thrown when trying to create a stream with a token with decimals greater than 18.

its not: 

https://github.com/sablier-labs/flow/blob/1e2c54392945a3ad8a2a598537849f78b5288788/src/libraries/Errors.sol#L19-L20

smol-ninja:
My bad. I wanted to point to the following

> [Errors.sol#L67-L68](https://cantina.xyz/code/c9980035-706a-4119-91fe-7ebddaf123e4/flow/src/libraries/Errors.sol#L67-L68): For clarity, consider updating the NatSpec of the SablierFlow_Unauthorized() error to
/// @notice Thrown when caller lacks authorization to perform an action.

src/libraries/Helpers.sol
andreivladbrg: I don’t think so—it’s more about how the lib is being used by the `SablierFlow` contract. It’s similar to how we added notes in `StreamingMath` from `lockup`:

https://github.com/sablier-labs/lockup/blob/8d9a1798c502db3137b7f6947755d246868bb36e/src/libraries/StreamingMath.sol#L37-L41

smol-ninja: The natspecs is about the helper function which does not have such requirement, as in it does not revert if decimal > 18. The function can be used with any decimal value and will always return some value. Therefore, I don't agree that this is a requirement.

1. `Assumptions` are fine as they are assuming things.
2. we should not say `must not` so there is nothing that can stop this function from accepting >18 decimal.

andreivladbrg:
> My bad. I wanted to point to the following
> 
> > [Errors.sol#L67-L68](https://cantina.xyz/code/c9980035-706a-4119-91fe-7ebddaf123e4/flow/src/libraries/Errors.sol#L67-L68): For clarity, consider updating the NatSpec of the SablierFlow_Unauthorized() error to
> > /// @notice Thrown when caller lacks authorization to perform an action.

I intentionally didn’t make this change (should’ve mentioned it) because we consistently refer to the caller as `msg.sender` throughout `ISablierFlow`. So I’d prefer to keep it as is for consistency.

src/interfaces/ISablierFlow.sol
```diff
@@ -111,8 +111,8 @@ interface ISablierFlow is
     /// @param streamId The stream ID for the query.
     function coveredDebtOf(uint256 streamId) external view returns (uint128 coveredDebt);
 
-    /// @notice Returns the time at which the total debt exceeds stream balance. If the total debt is less than
-    /// or equal to stream balance, it returns 0.
+    /// @notice Returns the time at which the total debt exceeds stream balance. If the total debt exceeds the stream
```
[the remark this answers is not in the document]
andreivladbrg: i personally prefer the current version (without "already")

src/libraries/Helpers.sol
andreivladbrg: How about: "The `decimals` must not be greater than 18, as values above that would cause an underflow."

andreivladbrg:
@smol-ninja pushed a commit to address your feedback, if it looks good, feel free to mark them as resolved.

smol-ninja:
> I intentionally didn’t make this change (should’ve mentioned it) because we consistently refer to the caller as msg.sender throughout ISablierFlow. So I’d prefer to keep it as is for consistency.

OK. Makes sense. 

> if it looks good, feel free to mark them as resolved

Sure. I will also wait for `setNativeToken` commit.

src/interfaces/ISablierFlow.sol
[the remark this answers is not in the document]
smol-ninja: OK. Not a big deal.

src/libraries/Helpers.sol
smol-ninja: I would prefer "should". "must" signals a struct requirement which should ideally be linked to a revert. From the function POV, it never reverts.

andreivladbrg: 1a8e10b30e21ff2c0e9e78bb3043ac2b33a1cd57

smol-ninja (approved):

PR #413, opened 2025-03-24
refactor: PRB feedback

PaulRBerg:


src/SablierFlow.sol
```diff
@@ -507,22 +507,20 @@ contract SablierFlow is
             return 0;
         }
 
+        // Check: if the rate per second is zero, skip the calculations.
```
smol-ninja, line 510: Shouldn't it be over the `if` branch?

src/abstracts/SablierFlowBase.sol
```diff
@@ -64,7 +64,8 @@ abstract contract SablierFlowBase is
         _;
     }
 
-    /// @dev Checks that `streamId` does not reference a paused stream.
+    /// @dev Checks that `streamId` does not reference a paused stream. Note that this implicitly checks
+    /// that the stream is not voided either.
```
smol-ninja, line 68: No it does not. A stream can be paused but not voided. I think you meant to add a similar statement over `notVoided` modifier.

smol-ninja (changes requested):
Some comments below.

src/SablierFlow.sol
PaulRBerg: Can you cover the full code so I can see what you mean?

smol-ninja:
```solidity
uint256 ratePerSecond = _streams[streamId].ratePerSecond.unwrap();

// Check: if the rate per second is zero, skip the calculations.
if (ratePerSecond == 0) {
    return 0;
}
```

PaulRBerg: idk it's up to you. I don't mind either way

smol-ninja: OK I will change it then.

src/abstracts/SablierFlowBase.sol
smol-ninja: https://github.com/sablier-labs/flow/pull/413/commits/9f980927312b5f65f35db25b2d73dfdac669b29e

src/SablierFlow.sol
smol-ninja: https://github.com/sablier-labs/flow/pull/413/commits/9f980927312b5f65f35db25b2d73dfdac669b29e

smol-ninja (approved):
LGTM. Feel free to merge.

Issue #414, opened 2025-03-26
Rename `aggregateBalance` to `aggregateAmount`


PaulRBerg:
The new name is better because it more clearly differentiates this variable from the ERC-20 balance, which can be greater than the `aggregateAmount`, hence leading to a surplus that can be recovered by the admin.

The current name `aggregateBalance` may be misunderstood as meaning the ERC20 `balanceOf` + something else.

PR #418, opened 2025-03-26
refactor: rename "aggregateBalance" to "aggregateAmount"

PaulRBerg:
- Closes https://github.com/sablier-labs/flow/issues/414
- Polishes the invariant tests

@andreivladbrg @smol-ninja for invariants, the descriptors should specify the negation instead of restating the invariant.

### DIFF
--- a/tests/invariant/Flow.t.sol
+++ b/tests/invariant/Flow.t.sol
@@ -58,81 +58,190 @@ contract Flow_Invariant_Test is Base_Test, StdInvariant {
     }
 
     /*//////////////////////////////////////////////////////////////////////////
-                                     INVARIANTS
+                              UNCONDITIONAL INVARIANTS
     //////////////////////////////////////////////////////////////////////////*/
 
-    /// @dev For a given token,
-    /// - the sum of all stream balances should equal the aggregate balance.
-    /// - token balance of the flow contract should be greater or equal to the sum of all stream balances
-    /// - sum of all stream balances should equal to the sum of all deposited amounts minus the sum of all refunded and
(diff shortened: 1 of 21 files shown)


tests/invariant/Flow.t.sol
```diff
@@ -58,321 +58,325 @@ contract Flow_Invariant_Test is Base_Test, StdInvariant {
...
+                              UNCONDITIONAL INVARIANTS
     //////////////////////////////////////////////////////////////////////////*/
 
-    /// @dev For a given token,
-    /// - the sum of all stream balances should equal the aggregate balance.
-    /// - token balance of the flow contract should be greater or equal to the sum of all stream balances
-    /// - sum of all stream balances should equal to the sum of all deposited amounts minus the sum of all refunded and
-    /// sum of all withdrawn.
-    function invariant_ContractBalanceStreamBalances() external view {
+    /// @dev Balances invariants:
```
smol-ninja, line 64: We keep a copy of invariants in `TECHNICAL-DOC` and make that in sync with the changes here so its easy to keep track of invariants.

Could you please update that as well?

smol-ninja (changes requested):
Looks good. Just two cents:

andreivladbrg (approved):
looks good 👍

smol-ninja (approved):

tests/invariant/Flow.t.sol
smol-ninja: Will handle it in https://github.com/sablier-labs/flow/issues/420

Issue #421, opened 2025-03-28
Remove DIAGRAMS and TECHNICAL-DOC


smol-ninja:
- [x] Remove [DIAGRAMS.md](https://github.com/sablier-labs/flow/blob/staging/DIAGRAMS.md)
- [x] Remove [TECHNICAL-DOC.md](https://github.com/sablier-labs/flow/blob/staging/TECHNICAL-DOC.md)
- [x] Move related content to the docs
- [x] Move the [Invariants section](https://github.com/sablier-labs/flow/blob/staging/TECHNICAL-DOC.md#invariants) into a new `README.md` within [`tests/invariant`](https://github.com/sablier-labs/flow/tree/staging/tests/invariant) directory.

PR #431, opened 2025-04-08
docs: refactor documentation

smol-ninja:
Closes https://github.com/sablier-labs/flow/issues/421

### Note

I am not sure where in [docs](https://docs.sablier.com/) we should move the precision section or should we even keep it at all? The precision content was helpful during the initial development of the protocol. And imo is no longer helpful at present and future stages. Let me know your thoughts.


andreivladbrg:
as stated before, concrete examples are a very good way to explain things to others — for any new reviewer of the flow code, they’d make things clearer, so we should keep them imo.
where exactly, not sure — will have to go through the flow docs

smol-ninja:
> concrete examples are a very good way to explain things to others

Yes, we can keep it. I just don't know where. Also, at this point, imo a good explanation on why we need scaling/descaling of amounts and why we deal with amounts in 18 decimals format are sufficient. That's all we have to answer.

smol-ninja:
@andreivladbrg can you please approve it if you have already reviewed it?

andreivladbrg:
> @andreivladbrg can you please approve the PR if you have already reviewed it?

haven't, will do it tomorrow, sorry for the delay

andreivladbrg (approved):
lgtm — the `README` file placed under `tests/invariant/` is more appropriate

there are no exact diagrams for the function calls from [here](https://github.com/sablier-labs/flow/blob/main/DIAGRAMS.md#function-calls)

i personally like it — should we create an issue for this one in `docs`, or are the other ones sufficient?

smol-ninja:
That diagram for the function calls is in the [docs here](https://docs.sablier.com/concepts/flow/statuses#functions-statuses-interaction). Its just split among 4 sub-sections but its exactly the same.

andreivladbrg:
> That diagram for the function calls is in the [docs here](https://docs.sablier.com/concepts/flow/statuses#functions-statuses-interaction). Its just split among 4 sub-sections but its exactly the same.

mb

PR #434, opened 2025-04-10
Remove remappings file

### LINKED ISSUE #433: Delete remappings.txt
As discussed in https://github.com/sablier-labs/flow/discussions/430, since foundry can handle autoremappings, delete `remappings.txt` file.

andreivladbrg:
Closes #433

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,6 +19,7 @@ typechain-types
 *.log
 .DS_Store
 .pnp.*
+bun.lockb
 deployments.md
 lcov.info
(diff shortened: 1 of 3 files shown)


smol-ninja (approved):
Thanks for the PR. I see that you only removed remappings reference from `Node.js` section but not from the `Git Submodules` section. Was it deliberate? If yes why? If no, feel free to update it and then merge this PR.

smol-ninja (approved):

Issue #419, opened 2025-03-26
Implement invariant for status transitions


PaulRBerg:
Similar to the invariant we have for Lockup:

https://github.com/sablier-labs/lockup/blob/557dc2c4f19fdd2ad9593de20624086f6e08e3d1/tests/invariant/Invariant.t.sol#L291

Patrick-Ehimen:
@PaulRBerg i be assigned this task?

PaulRBerg:
Hey @Patrick-Ehimen, thanks for your question and interest to continue making contributions to our repo! Appreciate this a lot.

Can you please tag @sablier-labs/evm instead? Doing so would also notify @smol-ninja and @andreivladbrg, which can assist you faster and better than I can.

smol-ninja:
Hey @Patrick-Ehimen, feel free to work on this issue. Let me know if you have any question about it.

Issue #417, opened 2025-03-26
Remove "isPaused" thingy


PaulRBerg:
The `isPaused` isn't helpful. It's outright confusing. This is the current implementation:

https://github.com/sablier-labs/flow/blob/3e76d3665fed248f85a0a3afa2afcb96d550760e/src/abstracts/SablierFlowBase.sol#L166-L168

This ~getter~thingy basically provides an alternative definition for what 'paused' means in SablierFlow.

'Paused' doesn't mean just one of the two enum statuses `PAUSED_SOLVENT` or `PAUSED_SOLVENT` — it also means the state of the RPS being zero.

That means that `isPaused` can return zero even when the stream's status is `VOIDED`.

There's really no need for this thingy. It's a very thin abstraction that does more harm than good by existing. It also complicates our invariants a lot.

PaulRBerg:
How do you determine if a stream is paused in the UI, @sablier-labs/frontend?

Do you use this getter `isPaused`, or do you check the status via `statusOf`?

razgraf:
Neither, we watch for the `Pause`/`Restart` events.

smol-ninja:
Are you proposing to rely on status than using getter function? If yes, then can we not use the same argument for `isVoided` getter as well?

PaulRBerg:
We can keep `isPaused` but it'd have to read the status and look for the `PAUSED_` statuses.

`isVoided` is OK because it doesn't introduce an alternative _meaning_ for the 'voided' concept.

My only beef is with the dual meaning of 'paused'.

smol-ninja:
Makes sense. Thanks for answering my question. I agree with you. Also, I just realized that classifying it a getter is also misleading since there is no state variables tracking paused status. A more appropriate getter is `getRatePerSecond` which is already there.

PaulRBerg:
You're welcome, and yeah good point!

PR #440, opened 2025-04-20
refactor: remove `isPaused` and add invariant for status transitions

smol-ninja:
### Changelog

- Closes https://github.com/sablier-labs/flow/issues/417
- Closes https://github.com/sablier-labs/flow/issues/419

### Note

I removed `isPaused` function from the contracts but left `notPaused` modifier intentionally. I think we should continue to use `paused` and `notPaused` terminology wherever it fits.

### DIFF
--- a/tests/integration/concrete/refund-and-pause/refundAndPause.tree
+++ b/tests/integration/concrete/refund-and-pause/refundAndPause.tree
@@ -16,7 +16,6 @@ RefundAndPause_Integration_Concrete_Test
          └── when caller sender
              ├── it should update the stream balance
              ├── it should perform the ERC20 transfer
-             ├── it should pause the stream
              ├── it should set rate per second to 0
              ├── it should update the snapshot debt
              ├── it should emit 1 {Transfer}, 1 {RefundFromFlowStream}, 1 {PauseFlowStream}, 1 {MetadataUpdate} events
--- a/tests/integration/concrete/refund-and-pause/refundAndPause.tree
+++ b/tests/integration/concrete/refund-and-pause/refundAndPause.tree
@@ -16,6 +16,7 @@ RefundAndPause_Integration_Concrete_Test
          └── when caller sender
              ├── it should update the stream balance
              ├── it should perform the ERC20 transfer
+             ├── it should pause the stream
              ├── it should set rate per second to 0
              ├── it should update the snapshot debt
              ├── it should emit 1 {Transfer}, 1 {RefundFromFlowStream}, 1 {PauseFlowStream}, 1 {MetadataUpdate} events
--- a/tests/invariant/README.md
+++ b/tests/invariant/README.md
@@ -23,18 +23,25 @@
 
 9. For any non-voided stream, if rps = 0 $\implies$ Flow.Status $\in$ {PAUSED_SOLVENT, PAUSED_INSOLVENT}
 
-10. For any non-pending stream, st $\le$ now.
+10. For any stream:
 
-11. For any non-voided stream, the snapshot time should never decrease
+    - If previous status is not pending, the current status should not be pending.
+    - If previous status is pending, the current status should neither be paused-solvent nor paused-insolvent.
+    - If previous status is paused-solvent, the current status should not be paused-insolvent.
+    - If previous status is voided, the current status should also be voided.
 
-12. For any non-voided stream, total streams = total debt + total withdrawals.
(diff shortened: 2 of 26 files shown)


tests/integration/concrete/refund-and-pause/refundAndPause.tree
```diff
@@ -16,7 +16,6 @@ RefundAndPause_Integration_Concrete_Test
          └── when caller sender
              ├── it should update the stream balance
              ├── it should perform the ERC20 transfer
-             ├── it should pause the stream
```
PaulRBerg, line 19: Wait, why remove this line? We should still check that the stream is paused via the status enum. Right?

PaulRBerg (approved):
I don't quite understand why we removed those branches in the BTT trees, but otherwise it LGTM

tests/integration/concrete/refund-and-pause/refundAndPause.tree
smol-ninja: It is still checking that the stream is paused by checking the `rps` value which is next line in the tree. This line was to check the `isPaused` state variable which I have removed.

In other words, when a stream is paused, it only sets the rps to 0. And that is still being checked.

PaulRBerg: Hoisted by our own petard?

I think we should still keep that branch because the function still says that it pauses something. The fact that the function now looks for `rpc == 0` to check the pausing simply means that we corrected the error. The BTT branch is still relevant, though.

Also, why not use the status enum?

smol-ninja: Yeah makes sense.

andreivladbrg: > The fact that the function now looks for rpc == 0 

it always did, not just now

i believe there is a contradiction between your point and the issue opened: why would we remove it in the first place then?  
before, `isPaused` would simply return true/false based on whether `rps == 0` or not. we had the invariant:  
`rps == 0 -> isPaused == true`  
so it was even redundant to test both IMO

```solidity
assertEq(rps == 0);
assertTrue(isPaused);
```

PaulRBerg: see emoji reactions

PaulRBerg: Sorry I am just having a little fun  😂

> i believe there is a contradiction between your point and the issue opened

There's no contradiction. As I said, the function still 'pauses' something, and so a branch that asserts that the function 'pauses' something makes complete sense. But what counts as 'pausing' is different now (and correct), whereas before it was confusing.

andreivladbrg: small thing, won't argue i am ok keeping it even tho i believe its not useful

tests/invariant/README.md
```diff
@@ -15,17 +15,20 @@
...
 
-9. For any stream, if rps $\gt$ 0 and no deposits are made $\implies \frac{d(ud)}{dt} \ge 0$
+9. For any non-voided stream, if rps = 0 $\implies$ Flow.Status $\in$ {PAUSED_SOLVENT, PAUSED_INSOLVENT}
 
-10. For any non-voided stream, if rps = 0 $\implies$ isPaused = true and Flow.Status $\in$ {PAUSED_SOLVENT,
-    PAUSED_INSOLVENT}
+10. For any stream:
+
+    - If previous status is not pending, the current status should not be pending.
+    - If previous status is pending, the current status should neither be paused-solvent nor paused-insolvent.
```
andreivladbrg, line 29: we've had this point before. why not prioritize what _it is_ instead of what _is not_?

i.e. why not say: "it should be either `STREAMING_SOLVENT || STREAMING_INSOLVENT`" ?

andreivladbrg (approved):
LGTM, just one comment:

tests/invariant/README.md
smol-ninja: Because here it makes sense to use `not pending`. It wont just be `STREAMING_SOLVENT || STREAMING_INSOLVENT` but also `PENDING || VOIDED` so we are talking about 4 out of 5 statuses.

andreivladbrg: oh, you are right.

Issue #444, opened 2025-05-06
Return refunded amount on `refundMax` call


gabrielstoica:
Hi!

Since Sablier Flow might be integrated into other protocols as well, returning the `refundedAmount` from the `refundMax` method would be a helpful addition. This would remove the need for an extra `refundableAmountOf` call to determine the amount being refunded to the stream sender.

Consider a scenario where a `SablierFlowManager` [contract](https://github.com/sablier-labs/evm-examples/blob/main/flow/FlowStreamManager.sol) is responsible for creating new streams. To enable guarded management operations (e.g., pausing or canceling), the manager contract acts as the sender (see a related discussion [here](https://github.com/sablier-labs/lockup/discussions/977)). Consequently, when a refund occurs, all funds are sent to the manager contract, which is then responsible for forwarding them to the original stream sender.

Currently, an additional `refundableAmountOf` call is required before invoking `refundMax` to know the actual amount that will be refunded by the manager contract.

This would also align with the behavior of `withdrawMax`.

PaulRBerg:
Great idea, thanks for the suggestion @gabrielstoica!

@sablier-labs/evm, let's include this change in the next release.

PR #445, opened 2025-05-06
feat: return refunded amount on `refundMax`

gabrielstoica:
### Changelog:
- Adds support for https://github.com/sablier-labs/flow/issues/444
- Updates tests for `refundMax` to check for the returned refunded amount
- Corrects some _checks_ comments

### DIFF
--- a/src/SablierFlow.sol
+++ b/src/SablierFlow.sol
@@ -369,11 +369,12 @@ contract SablierFlow is
         notNull(streamId)
         onlySender(streamId)
         updateMetadata(streamId)
+        returns (uint128 refundedAmount)
     {
-        uint128 refundableAmount = _refundableAmountOf(streamId);
+        refundedAmount = _refundableAmountOf(streamId);
 
         // Checks, Effects, and Interactions: make the refund.
-        _refund(streamId, refundableAmount);
+        _refund(streamId, refundedAmount);
     }
 
     /// @inheritdoc ISablierFlow
@@ -509,7 +510,7 @@ contract SablierFlow is
 
         uint256 ratePerSecond = _streams[streamId].ratePerSecond.unwrap();
 
-        // Check: if the rate per second is zero, skip the calculations.
+        // Check: the rate per second is zero, skip the calculations.
         if (ratePerSecond == 0) {
             return 0;
         }
@@ -623,7 +624,7 @@ contract SablierFlow is
 
         uint40 blockTimestamp = uint40(block.timestamp);
 
-        // Check: if the start time is in the future, the rate per second is not zero.
+        // Check: the start time is in the future and the rate per second is not zero.
         if (startTime > blockTimestamp && ratePerSecond.unwrap() == 0) {
             revert Errors.SablierFlow_CreateRatePerSecondZero();
         }
--- a/src/SablierFlow.sol
+++ b/src/SablierFlow.sol
@@ -510,7 +510,7 @@ contract SablierFlow is
 
         uint256 ratePerSecond = _streams[streamId].ratePerSecond.unwrap();
 
-        // Check: the rate per second is zero, skip the calculations.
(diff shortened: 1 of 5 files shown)


smol-ninja (changes requested):
Thank you for the PR @gabrielstoica. I agree with [your rationale](https://github.com/sablier-labs/flow/issues/444) of returning the amount in case of `refundMax`. The implementation looks good to me. There is a bit of feedback regarding the comments. I will wait for your reply before merging this PR.

gabrielstoica:
Hey @smol-ninja, no worries!

I think there was a similar situation in the past on the "Check:" word followed by an "if". Here's [a reference](https://github.com/sablier-labs/lockup/pull/1054#discussion_r1791843343).

Let me know if something has changed and I can easily update everything.

smol-ninja:
That reference from the past still holds true (that the case of revert).

> Check: `msg.sender` is neither the recipient nor approved third party

implies `msg.sender` must either be recipient or an approved third party (otherwise it would revert).

vs

> Check: the start time is in the future and the rate per second is not zero.

implies that the start time must be in the future and rps must be 0, otherwise it would revert.

vs

> Check: if the start time is in the future, rate per second is not zero.

implies that "only if" start time is in future, rps must not be zero, otherwise it would revert. But it does not say anything about start time _not in_ the future and in those cases no checks are required.

Similarly,

> Check: `msg.sender` has neither the {RoleAdminable.FEE_COLLECTOR_ROLE} role nor is the contract admin, then `feeRecipient` must be the admin address

There are two statements: first one defines conditions, second one defines action. In this case, I am not sure if its clear without `if` preposition.

---------

We still don't use `if` if there is only one statement and the hidden action is to revert. But when two or more statements are required to trigger the hidden revert action, `if` would make it extra clear.

Thanks for opening this discussion. I would love to know what you think and if you have any feedback on my thought process.

gabrielstoica:
That makes sense - thanks for the thoughtful rationale.

As I take a second look at everything, my take is that sticking to the rules is the right approach. In this case, the check definitions should be adjusted to sound correct, for example by using _**then**_ to indicate a consequence or result.

from:
> Check: if the rate per second is zero, skip the calculations.

to:
> **Check: the rate per second is zero, then skip the calculations.**

---
from:
> Check: if the start time is in the future, the rate per second is not zero.

to:
> **Check: the start time is in the future, then the rate per second must not be zero.**

---
from:
> Check: if `msg.sender` has neither the {RoleAdminable.FEE_COLLECTOR_ROLE} role nor is the contract admin, then `feeRecipient` must be the admin address.

to:
> **Check: `msg.sender` has neither the {RoleAdminable.FEE_COLLECTOR_ROLE} role nor is the contract admin, then `feeRecipient` must be the admin address.**

This way, the _check rule_ is validated even without an explicit _if_ formula, and the condition-action rationale remains clear. 
Let me know your thoughts.

smol-ninja:
> thanks for the thoughtful rationale

You're welcome and likewise.

> my take is that sticking to the rules is the right approach

I do agree with this as long as there is no loss of clarity in the explanation. 

Though using `then` instead of `if` conveys the same message, I don't think its grammatically sound. `if` makes the statement read like a condition-action. On the other hand, when you are using `then` (by dropping `if`), it sounds more like two statements connected through `then`. 

Here is another example,

1. "If Columbus sails west, Vasco da Gama  sails east": clear condition–action relationship.

2. "Columbus sails west, then Vasco da Gama  sails east": two statements connected by `then`.

Though missing `then` in `If Columbus sails west, then Vasco da Gama  sails east` would not make it less sound as much as missing `if` does.

gabrielstoica:
That makes sense. I'm happy to update everything to match this.

smol-ninja (approved):
Thank you for the PR and making the changes. Looks good to me now.

pikachu

PR #451, opened 2025-06-12
refactor: use comptroller

andreivladbrg:


andreivladbrg:
@smol-ninja tentatively requested a review as it's pretty much done,  i just commented the comptroller handler target in invariants as it would break the tests

not sure why, although it's the same as in `lockup`, the only difference is the multiple tokens being used for inv tests in flow. 
will further debug tomorrow as it's a small thing and won't affect the review

tests/invariant/handlers/FlowComptrollerHandler.sol
```diff
@@ -45,8 +36,9 @@ contract FlowComptrollerHandler is BaseHandler {
     {
         vm.assume(currentToken.balanceOf(address(flow)) > flow.aggregateAmount(currentToken));
 
-        setMsgSender(address(flow.comptroller()));
+        ISablierComptroller comptroller = flow.comptroller();
 
-        flow.recover(currentToken, address(comptroller));
```
smol-ninja: The issue was here. This contract inherits from Base but does not call `setUp` (which is called in `Flow_Invariant_Test` and not here) and therefore `comptroller` was being set as null address sometimes. And thats why it was failing.

To debug it, I looked at the error in invariant test and compared it with the selectors from `forge selectors list` to understand where its throwing error.

andreivladbrg: https://github.com/sablier-labs/evm-utils/issues/36

Issue #303, opened 2024-10-13
tests for values of rps near `uint128` max


smol-ninja:
Many of [Cantina findings](https://cantina.xyz/code/99ae802b-f05c-4e36-a1d1-240d5146649c/findings/10) were due to the fact that there are no concrete tests that can test the Flow system for max possible values for `rps`. Therefore, brainstorm potential solutions to achieve this. 

The advantage is clear: Flow protocol would be tested for edge cases.

This can also be done in conjecture with https://github.com/sablier-labs/flow/issues/272.

PaulRBerg:
Should we still keep this open @sablier-labs/evm?

andreivladbrg:
i think this test is suffice

https://github.com/sablier-labs/flow/blob/a96bd253d926a29faf504c7571901d66efeb78ab/tests/integration/fuzz/withdrawMultiple.t.sol#L63

right, @smol-ninja?

otherwise values closer to `uint128` will overflow very quickly

smol-ninja:
My first thought was to close this issue but we do not have tests that test for `rps`> `2e18` which can be realistic imo. So, I would like to keep this issue open and write a test for rps in the range `(2e18, max)`. 

> otherwise values closer to uint128 will overflow very quickly

Yes thats fine and it should revert in those cases.

andreivladbrg:
alright - agree

PaulRBerg:
more testing makes me happy

good idea to keep open

PR #459, opened 2025-08-08
test(fuzz): multiple withdrawals with large rps

smol-ninja:
Closes https://github.com/sablier-labs/flow/issues/303

I have removed `NoDelay` from the names as I don't think there is any delay anymore.


smol-ninja:
Sorry I requested too early @andreivladbrg but now its ready for review.

PR #465, opened 2025-09-09
test: use vm.getBlockTimestamp()

PaulRBerg:
See context in https://github.com/sablier-labs/evm-utils/issues/62


smol-ninja (approved):
I made some changes so it looks good to me now.

PR #474, opened 2025-10-21
test: remove unneeded solady import in test utils

andreivladbrg:


smol-ninja:
We should remove `solady` from `package.json` as well since its not used anywhere else. Also, can you bump the patch version too after this PR is merged?

smol-ninja (approved):

PR #477, opened 2025-11-04
docs: add withdraw requirement about msg.value

andreivladbrg:


smol-ninja:
Can you add it to `monorepo` code as well?

smol-ninja (approved):

